PR #8, opened 2022-12-12
Add TypedField for block based content management system

alexander-schranz:
We need some kind of metadata for type based blocks. The TypedField will allow us this:

```php
    'blocks' => new Field\TypedField('blocks', 'type', [
        'text' => [
            'title' => new Field\TextField('title'),
            'description' => new Field\TextField('description'),
            'media' => new Field\IntegerField('media', multiple: true),
        ],
        'embed' => [
            'title' => new Field\TextField('title'),
            'media' => new Field\TextField('media'),
        ],
    ], multiple: true),
```

### DIFF
--- a/src/SEAL/README.md
+++ b/src/SEAL/README.md
@@ -22,6 +22,24 @@ $document = [
     'id' => '1',
     'title' => 'New Blog',
     'article' => '<article><h2>Some Subtitle</h2><p>A html field with some content</p></article>',
+    'blocks' => [
+        [
+            'type' => 'text',
+            'title' => 'Titel',
+            'description' => '<p>Description</p>',
+            'media' => ['id' => 1, 'displayOption' => 'top'],
+        ],
+        [
+            'type' => 'text',
+            'title' => 'Titel 2',
+            'description' => '<p>Description 2</p>',
+        ],
+        [
+            'type' => 'embed',
+            'title' => 'Video',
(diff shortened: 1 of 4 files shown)


src/SEAL/README.md
```diff
@@ -22,6 +22,24 @@ $document = [
     'id' => '1',
     'title' => 'New Blog',
     'article' => '<article><h2>Some Subtitle</h2><p>A html field with some content</p></article>',
+    'blocks' => [
+        [
+            'type' => 'text',
+            'title' => 'Titel',
+            'description' => '<p>Description</p>',
+            'media' => ['id' => 1, 'displayOption' => 'top'],
```
alexander-schranz, line 30: The tricky part is supporting this kind of things here. A property which need to be handled differently based on the type.

src/SEAL/README.md
```diff
@@ -22,6 +22,24 @@ $document = [
...
+        ],
+        [
+            'type' => 'text',
+            'title' => 'Titel 2',
+            'description' => '<p>Description 2</p>',
+        ],
+        [
+            'type' => 'embed',
+            'title' => 'Video',
+            'media' => 'https://www.youtube.com/watch?v=iYM2zFP3Zn0',
```
alexander-schranz, line 40: The tricky part is supporting this kind of things here. A property which need to be handled differently based on the type.

PR #12, opened 2022-12-26
Split seal package into Schema and DSL

wachterjohannes:
@alexander-schranz what do you think? this would help to create packages which only use the DSL or the Schema.

### DIFF
--- a/config.subsplit-publish.json
+++ b/config.subsplit-publish.json
@@ -6,6 +6,16 @@
             "target": "git@github.com:schranz-search/seal.git"
         },
         {
+            "name": "DSL",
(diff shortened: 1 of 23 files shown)


alexander-schranz:
Would not create a `DSL` namespace. Did go with Search as I like that more so would keep that, as providing classes and a builder is I think not a DSL like DQL.

PS: the split of into `Search`/`DSL` does not make sense as it has a requirement to the Adapter via the SearchBuilder so the only thing which currently can be put into an own package would be the schema, which I think does not make sense if you already have dependency on the Search.

PR #50, opened 2023-01-06
Add Search Condition to Search Engines

alexander-schranz:
The `SearchCondition` is the most used Condition it searches a query inside the provided indexes. And so is build to great traditional searches.


packages/seal-elasticsearch-adapter/ElasticsearchConnection.php
```diff
@@ -119,6 +120,8 @@ public function search(Search $search): Result
         foreach ($search->filters as $filter) {
             if ($filter instanceof IdentifierCondition) {
                 $query['ids']['values'][] = $filter->identifier;
+            } elseif ($filter instanceof SearchCondition) {
+                $query['query_string']['query'] = $filter->query;
```
alexander-schranz, line 124: @wachterjohannes elasticsearch is a little bit confusing here not sure what the best case is for search that query.

wachterjohannes: would keep it like that - i would say it is the "normal" search! but the better option would be `match` but there you have to set a specific field

PR #79, opened 2023-01-15
Add support for Zinc

alexander-schranz:
Zinc would be another alternative to Elasticsearch. The server is available here: https://github.com/zinclabs/zinc

Sadly we could not yet the docker compose get running which is required for local development and github CI:

 - [x] https://github.com/zinclabs/zinc/issues/710
 - [ ] https://github.com/zinclabs/zinc/issues/708


alexander-schranz:
I'm closing this as I think @zincsearch Team focus seems to be now on @openobserve. Which is more on logs, metrics as for app search.

Issue #154, opened 2023-04-14
[Docs] Clarification of terminology in the "Getting Started" section of the documentation is missing


froschdesign:
The ["Getting Started" section](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst) of the documentation uses terms that are not described, but the entire section is based on this terminology:

- Schema
- Engine
- Search Engine
- Indexes
- Documents

The terms are used in the headlines and content:

- [Configure Schema](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#configure-schema)
- [Configure Engine](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#configure-schema)
- [Prepare Search Engine](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#prepare-search-engine)
- [Create Indexes](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#create-indexes)
- [Add Documents](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#add-documents)
- [Search Documents](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#search-documents)
- [Filter Documents](https://github.com/schranz-search/schranz-search/blob/0.1/docs/getting_started/index.rst#filter-documents)

Additionally, the use of "engine" and "search engine" without explanation is very confusing or even unclear.

--- 

One description can be found: 

> The `Engine` is the main class which will be used to communicate with the search engine.

Therefore, I suggest adding a list of terms with descriptions at the top of the page.

alexander-schranz:
@froschdesign  Thx for the feedback. Maybe it was not a good choice to name that class `Engine` 🤔, but did not come up with a better name yet.

Beside describing the terms also, do you think it would avoid confusion if we use instead of `Search engines` the terms `Search services`, or what other word could use here to not use Engine here also 🤔 .

alexander-schranz:
I added a glossary to the first page of the docs to make things more clear as part of the diagram and structure docs: https://schranz-search.github.io/schranz-search/

PR #161, opened 2023-04-17
Rename search engines in docs to search services

alexander-schranz:
part of #154. Not 100% sure if this change is good so currently just a draft.

### DIFF
--- a/docs/getting_started/index.rst
+++ b/docs/getting_started/index.rst
@@ -46,7 +46,7 @@ integration of the package or the ``Standalone`` version.
 
              composer require schranz-search/spiral-bridge
 
-| The project provides adapters to different search engines, atleast one is required.
+| The project provides adapters to different search services, atleast one is required.
 | Choose the one which fits your needs best:
 
 .. tabs::
@@ -194,7 +194,7 @@ Configure Engine
 ----------------
 
 In the next step we will create the engine which will be use our created ``Schema``.
-The ``Engine`` is the main class which will be used to communicate with the search engine.
+The ``Engine`` is the main class which will be used to communicate with the search service.
 It requires an instance of the ``Adapter`` which we did install before.
 
 .. tabs::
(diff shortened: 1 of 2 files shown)


docs/getting_started/index.rst
```diff
@@ -46,7 +46,7 @@ integration of the package or the ``Standalone`` version.
 
              composer require schranz-search/spiral-bridge
 
-| The project provides adapters to different search engines, atleast one is required.
+| The project provides adapters to different search services, atleast one is required.
```
froschdesign, line 49: The problem is that various providers call their products "search engines" themselves:

> Meilisearch is a flexible and powerful user-focused search engine…

https://www.meilisearch.com

> Elasticsearch is a distributed, RESTful search and analytics engine…

https://www.elastic.co/elasticsearch/

> Free and Open, Distributed, RESTful Search Engine

https://github.com/elastic/elasticsearch

> Typesense is a modern, privacy-friendly, open source search engine…

https://typesense.org

And also Wikipedia:

> Elasticsearch is a search engine…

https://en.wikipedia.org/wiki/Elasticsearch

---

But Wikipedia uses another term: [Search engine software](https://en.wikipedia.org/wiki/List_of_search_engine_software)

alexander-schranz: Thx for the feedback. Maybe adding some `abbr` can avoid confusing between terms: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-abbr

froschdesign: I would prefer unambiguous and clearly readable names. Maybe something like these:

- Search engine software
- Search client
- …

vs.

- SEAL engine
- Abstract engine
- Search connector
- …

If the engine class is based on a design pattern, then this name could also be used / added.

:thinking:

alexander-schranz:
close in favor of #163

docs/getting_started/index.rst
alexander-schranz: Didn't wanted to use `client` as client is mostly provided by specific adapter classes and could be confused with that one easily, so in the PHP autocomplete there is not much conflicting with the Engine class, so will currently keep that one.

PR #166, opened 2023-04-20
Add Mezzio Integration

### LINKED ISSUE #164: [Mezzio] Integration Module
It would be nice to support also Mezzio framework. Currently I did only have a short look how this could be integrated, and failed to understand how I can create services based on configuration. I know how to get config inside a factory but not how the config can control that one or multipl services are created.

Every service seems require a factory be configured at the time where we don't have the config (e.g.: https://github.com/schranz-templating/templating/blob/eb28f45eabec42a5738b999030ab8e9e049d4fd7/src/Integration/Mezzio/Latte/ConfigProvider.php#L31-L35). So we can do `Engine::class -> MezzioEngineFactory::class` but not sure how to handle this if we want to create based on configuration multiple instances of the `Engine` like we are doing it here in the other frameworks:

 - https://github.com/schranz-search/schranz-search/blob/b5fe051db7618bee9a6ea27674467d5e41f76646/integrations/laravel/src/SearchProvider.php#L79
 - https://github.com/schranz-search/schranz-search/blob/b5fe051db7618bee9a6ea27674467d5e41f76646/integrations/spiral/src/Bootloader/SearchBootloader.php#L84
 - https://github.com/schranz-search/schranz-search/blob/b5fe051db7618bee9a6ea27674467d5e41f76646/integrations/symfony/src/SearchBundle.php#L70

I also did have a look at the `Doctrine` integration but not understand how it was handled there that multipl instances of entity manager can exist 🤔 .

alexander-schranz:
TODO:

 - [x] Skeleton
 - [x] Init Module 
 - [x] Adapter Integration (ReadWrite / Multi Adapter not working) (maybe fix via SEALContainer could solve this see also [here](https://github.com/schranz-search/schranz-search/issues/164#issuecomment-1516992056))
 - [x] Schema Integration 
 - [x] Engine Integration
 - [x] EngineRegistry Integration 
 - [x] Prefix config
 - [x] Add Lint
 - [x] Add Tests
 - [x] Create & Drop CLI Commands
 - [x] Add CLI Tests
 - [x] Update Docs
 - [x] Publish Package
 
fixes #164


integrations/mezzio/src/Service/AdapterFactoryAbstractFactory.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function __invoke(ContainerInterface $container, string $className): AdapterFactoryInterface
+    {
+        if (
+            ReadWriteAdapterFactory::class === $className
+            || MultiAdapterFactory::class === $className
+        ) {
+            return new $className(
+                $container,
+                'schranz_search.adapter.', // FIXME this currently does not work as the adapter instance is not a service
```
alexander-schranz: In the current setup this is not working as the `EngineRegistryFactory` is creating the `Adapter` and we can not access it via the `$container` here.

alexander-schranz: This is fixed by creating an own `SealContainer` class which register our services

PR #167, opened 2023-04-21
Refactored Spiral Bridge and Update README

butschster:
Hello @alexander-schranz,

I have made some changes to the Spiral Bridge package, and I would like to contribute these changes to the project. Specifically, I have refactored the code of the Bridge to make it more readable and updated the README file to provide more accurate information about the package.

I believe these changes will improve the usability and maintainability of the package, and I hope you find them useful.

Thank you for providing the integration with the Spiral Framework!


integrations/spiral/README.md
```diff
@@ -69,21 +83,26 @@ return [
 ];
 ```
 
-A more complex configuration can be here found:
+> **Warning**
+> When defining the `engines` section, make sure to use the `default` key to set the default engine. This default engine
+> can then be accessed through the `Schranz\Search\SEAL\Engine` class. If you don't explicitly define a default engine,
+> the first engine listed will be used as the default.
```
alexander-schranz: This is not the expected case of the configuration

integrations/spiral/README.md
```diff
@@ -23,42 +24,55 @@ Use [composer](https://getcomposer.org/) for install the package:
...
 
-The following code shows how to configure the package:
+At first, you need to create a configuration file `app/config/schranz_search.php` that helps you to configure search
+schemas and engines. Below you can find an example of a simple configuration file:
 
 ```php
 <?php
-
 // app/config/schranz_search.php
+declare(strict_types=1);
```
alexander-schranz: unnecessary and so can be removed

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -35,8 +34,21 @@
  */
 final class SearchBootloader extends Bootloader
 {
-    protected const DEPENDENCIES = [
-        ConsoleBootloader::class,
+    protected const SINGLETONS = [
+        AdapterFactory::class => [self::class, 'createAdapterFactory'],
+        EngineRegistry::class => [self::class, 'createEngineRegistry'],
+        Engine::class => [self::class, 'defaultEngine'],
```
alexander-schranz: No this is not the way how it should work and so it should be kept that way, as before that always a default engine exist. As it should be handled in all integrations the same way.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
         $engineServices = [];
         foreach ($engines as $name => $engineConfig) {
-            $adapterServiceId = 'schranz_search.adapter.' . $name;
-            $engineServiceId = 'schranz_search.engine.' . $name;
-            $schemaLoaderServiceId = 'schranz_search.schema_loader.' . $name;
-            $schemaId = 'schranz_search.schema.' . $name;
+            $adapterServiceId = 'schranz_search.adapter.'.$name;
+            $engineServiceId = 'schranz_search.engine.'.$name;
+            $schemaLoaderServiceId = 'schranz_search.schema_loader.'.$name;
+            $schemaId = 'schranz_search.schema.'.$name;
```
alexander-schranz: Keep the codestyle here.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
             $adapterDsn = $engineConfig['adapter'];
             $dirs = $engineSchemaDirs[$name] ?? [];
 
-            $container->bindSingleton($adapterServiceId, function (Container $container) use ($adapterDsn) {
-                /** @var AdapterFactory $factory */
-                $factory = $container->get(AdapterFactory::class);
-
-                return $factory->createAdapter($adapterDsn);
-            });
+            $container->bindSingleton($adapterServiceId, $adapter = $adapterFactory->createAdapter($adapterDsn));
```
alexander-schranz: This looks very unexpected, if that would be the case we would not even need `$container->bindSingleton` as this way we not even allow service decoration as `$adapter` is used instead of the service inside the container.

Same for the other cases. I would not go this way as from dependency injection point of view this looks not correctly for me.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
-                $adapter = $container->get($adapterServiceId);
-                /** @var Schema $schema */
-                $schema = $container->get($schemaId);
-
-                return new Engine($adapter, $schema);
-            });
-
-            if ('default' === $name || (!isset($engines['default']) && !$container->has(Engine::class))) {
-                $container->bind(Engine::class, $engineServiceId);
-            }
```
alexander-schranz: This need to be kept as written above, all integrations work the same way.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
+            $engineServices[$name] = $engine;
         }
 
-        $container->bindSingleton(EngineRegistry::class, function (Container $container) use ($engineServices) {
-            $engines = [];
-
-            foreach ($engineServices as $name => $engineServiceId) {
-                /** @var Engine $engine */
-                $engine = $container->get($engineServiceId);
```
alexander-schranz, line 130: this should be kept also as written above we should go over container and not some local variables else service decorations and other kind of things don't work like expected.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
-                $engines[$name] = $engine;
-            }
-
-            return new EngineRegistry($engines);
-        });
+        return new EngineRegistry($engineServices);
     }
 
-    private function createAdapterFactories(Container $container): void
+    private function defaultEngine(EngineRegistry $registry): Engine
```
alexander-schranz: Not necessary as this should as written be handled differently if there is no `default` define still a default engine is created, all integration should work the same.

integrations/spiral/README.md
```diff
@@ -124,49 +143,78 @@ return [
 ];
 ```
 
-The default engine is available as `Engine`:
+## Usage
```
alexander-schranz: basic example should be keept in all integrations the same way and as small as possible.

This part would be more some kind of a cookbook how to index `Spiral` `Models` into the index.

integrations/spiral/src/Bootloader/SearchBootloader.php
```diff
@@ -82,144 +95,82 @@ public function boot(ConsoleBootloader $console, Container $container): void
...
+            /** @var AdapterFactoryInterface $adapterFactory */
+            $adapterFactory = $container->get($adapterServiceId);
 
-                $factories[$name] = $adapterFactory;
-            }
+            $factories[$name] = $adapterFactory;
+        }
 
-            return new AdapterFactory($factories);
-        });
```
alexander-schranz: Looks unexpected for me to throw this away as this avoids that a service can be decorated over the container and from my point of view that is the whole part why using a dependency injection container to allow manipulation of the services.

integrations/spiral/src/Config/SearchConfig.php
```diff
@@ -8,6 +8,16 @@
...
+ * @psalm-type TEngine = array{
+ *     adapter: string,
+ * }
+ * @psalm-type TEngines = array<string, TEngine>
+ * @psalm-type TSchema = array{
+ *     dir: string,
+ *     engine?: string,
+ * }
+ *
+ * @psalm-type TSchemas = array<non-empty-string, TSchema>
```
alexander-schranz: we are not using psalm in this project so PHPStan types should be used here.

integrations/spiral/src/Console/IndexDropCommand.php
```diff
@@ -42,13 +43,13 @@ public function __invoke(EngineRegistry $engineRegistry): int
...
             if ($this->indexName) {
-                $this->line('Drop search index "' . $this->indexName . '" of "' . $name . '" ...');
+                $this->line(\sprintf('Drop search index "%s" of "%s" ...', $this->indexName, $name));
                 $engine->dropIndex($this->indexName);
 
                 continue;
             }
 
-            $this->line('Drop search indexes of "' . $name . '" ...');
+            $this->line(\sprintf('Drop search indexes of "%s" ...', $name));
```
alexander-schranz: to be consistent to the whole code base we should keep it as it was before.

alexander-schranz (changes requested):
@butschster Thank you for taking time to have a look at this project. And thank you for your feeback.

I have added some comments mostly aboud remove tin `$container->get` feels wrong from my point of view in part of `DependencyInjection` as I think services should be fetched from the container not from a local variables to make services decoratable over the Spiral Dependency Injection container. This way we just expose a fixed instance but don't allow that somebody decorate it.

integrations/spiral/README.md
butschster: Second part or first one?

integrations/spiral/README.md
butschster: It's not about index spiral models, there is an abstract code, which shows only more understandable example, what to do with $engine object.

integrations/spiral/README.md
butschster: As I can see in your code `Engine` will be associated with `default` engine if it exists, or with first engine otherwise

alexander-schranz (approved):
Really thank you for testing and providing a pull request. The README I will rework in seperate PR, to tackle all READMEs the same way, thank you there for the feedback.

PR #222, opened 2023-08-05
Add support for Loupe PHP

alexander-schranz:
This allows to index documents into Loupe PHP a SQLite based PHP only search engine by @toflar. 

 - https://github.com/loupe-php/loupe
 
# Requires

 - [x] https://github.com/loupe-php/loupe/pull/10 (currently solved via reflection)
 - [x] https://github.com/loupe-php/loupe/pull/12
 - [x] https://github.com/loupe-php/loupe/issues/13
 - [x] https://github.com/loupe-php/loupe/pull/14
 - [ ] https://github.com/loupe-php/loupe/issues/15 (currently solved via LoupeHelper)

# TODO

 - [x] SchemaManagerTest
   - [x] testSimpleSchema
   - [x] testComplexSchema
 - [x] AdapterTest
   - [x] testIndex
   - [x] testSchema
   - [x] testDocument
 - [x] IndexerTest
   - [x] testSaveDeleteIdentifierCondition
 - [x] SearcherTest
   - [ ] ~~testFindMultipleIndexes~~ (currently already only supported by Elasticsearch and Opensearch) so MultiIndex search will maybe be removed from the abstraction
   - [x] testSearchCondition
   - [x] testNoneSearchableFields
   - [x] testLimitAndOffset
   - [ ] TODO add test case for none paginatable limit offset values ([see here](https://github.com/loupe-php/loupe/issues/13) )
   - [ ] TODO add test case when adding fields after first document ([see here](https://github.com/loupe-php/loupe/issues/12) )
   - [x] testEqualCondition
   - [x] testMultiEqualCondition
   - [x] testNotEqualCondition([see here](https://github.com/loupe-php/loupe/pull/14))
   - [x] testGreaterThanCondition
   - [x] testGreaterThanEqualCondition
   - [x] testLessThanCondition
   - [x] testLessThanEqualCondition
   - [x] testSortByAsc
   - [x] testSortByDesc
 - [x] Integrations
   - [x] Symfony
   - [x] Laravel
   - [x] Spiral
   - [x] Mezzio
   - [x] Yii
 - [x] Documentation
   - [x] Supported Engines
   - [x] Graphic
   - [x] OG Image
   
Fixed Errors in:

 - https://github.com/loupe-php/loupe/pull/10
 - https://github.com/loupe-php/loupe/pull/12
 - https://github.com/loupe-php/loupe/pull/14

Bildschirmfoto 2023-08-06 um 01 17 19

So with the above PRs the Loupe adapter finish all currently implemented tests.


packages/seal-loupe-adapter/src/LoupeHelper.php
```diff
@@ -0,0 +1,126 @@
...
+
+use Loupe\Loupe\Configuration;
+use Loupe\Loupe\Loupe;
+use Loupe\Loupe\LoupeFactory;
+use Schranz\Search\SEAL\Schema\Index;
+
+/**
+ * @experimental
+ */
+final class LoupeHelper
```
alexander-schranz, line 24: @toflar What is a little bit different to other search engines is that at creating the `Loupe` instance (what in other cases is the `client`) the whole configuration searchablefields, filterablefields, .. is required where in other cases this I only need to know about in the SchemeManager where I call createIndex method not in cases of searching or indexing documents.

For this I created this Helper class which could also be called `LoupeManager`. It even `serialize` the configuration into a file this way it is more strict as changing updating the configuration can only be done like in other engine via the SchemaManager::createIndex method and so it is more consistent through the whole library and so without calling that the changes apply no effects.

For my usecases I maybe will be changing it when I got more time.

But it could also be part of Loupe, a none SEAL related `LoupeManager` could look like this. What do you think about it?

```php
class LoupeManager
{
    /**
     * @var Loupe[]
     */
    private array $loupe = [];

    public function __construct(
        private readonly LoupeFactory $loupeFactory,
        private string $directory,
    ) {
    }
    
    public function getLoupe(string $index): Loupe
    {
        if (!isset($this->loupe[$index->name])) {
            $configurationFile = $this->directory . $index . '.loup';

            if (!\file_exists($configurationFile)) {
                throw new \LogicException('Index need to be created before accessing Loupe instance.');
            }

            /** @var string $configurationContent */
            $configurationContent = \file_get_contents($configurationFile);

            /** @var Configuration $configuration */
            $configuration = \unserialize($configurationContent);
            
            $this->loupe[$index->name] = $this->loupeFactory->create($this->directory . $index . '.db', $configuration);
        }

        return $this->loupe[$index->name];
    }
    
    public function existIndex(string $index): bool
    {
        return file_exists(($this->directory . $index . '.db');
    }
    
    public function dropIndex(string $index): void
    {
         unlink($this->directory . $index . '.db', $this->directory . $alias . '.db');
         unlink($this->directory . $index . '.loupe', $this->directory . $alias . '.loupe');
    }
    
    public function createIndex(string $index, Configuration $configuration): void
    {
        $configurationContent = \serialize($configuration);
        
        file_put_contents($this->directory . $index . '.loupe', $this->directory . $alias . '.loupe', $configurationContent);
        touch($this->directory . $index . '.db');
    }
    
    public function alias(string $index, string $alias): void
    {
         // aliases via symlinks for zerodowntime reindexing
         symlink($this->directory . $index . '.db', $this->directory . $alias . '.db');
         symlink($this->directory . $index . '.loupe', $this->directory . $alias . '.loupe');
    }
    
    public function removeAlias(string $alias): void
    {
         unlink($this->directory . $alias . '.db');
         unlink($this->directory . $alias . '.loupe');
    }
}
```

Maybe saving the `configuration` in the DB instead of a file would be better.

Did put this also in an issue https://github.com/loupe-php/loupe/issues/15

packages/seal-loupe-adapter/composer.json
```diff
@@ -0,0 +1,85 @@
...
+    "authors": [
+        {
+            "name": "Alexander Schranz",
+            "email": "alexander@sulu.io"
+        }
+    ],
+    "require": {
+        "php": "^8.1",
+        "schranz-search/seal": "^0.1",
+        "loupe/loupe": "dev-enhancement/fixes-together as 0.1.1",
```
alexander-schranz: TODO revert this

packages/seal-loupe-adapter/src/LoupeHelper.php
Toflar: Answered in the ticket :)

PR #224, opened 2023-08-07
Check what happens when first element is empty

alexander-schranz:
This tests what happens if the first given document has not all fields: https://github.com/loupe-php/loupe/pull/12#issuecomment-1667440576

Another check can be found here: https://github.com/schranz-search/schranz-search/pull/225


alexander-schranz:
Closing this as this was just a test show what happens if the first document does not contain all fields: https://github.com/loupe-php/loupe/pull/12#issuecomment-1668395330

PR #233, opened 2023-08-24
Fix issue with new redis version

alexander-schranz:
Keep it like the other docker images on the same version.


packages/seal-redisearch-adapter/src/RediSearchSchemaManager.php
```diff
@@ -31,7 +31,7 @@ public function existIndex(Index $index): bool
         try {
             $indexInfo = $this->client->rawCommand('FT.INFO', $index->name);
         } catch (\RedisException $e) {
-            if ('Unknown Index name' !== $e->getMessage()) {
+            if ('unknown index name' !== \strtolower($e->getMessage())) {
```
alexander-schranz, line 34: This seems now be returned differently based on the redis server version so we need to strtolower now.

PR #245, opened 2023-08-29
Add conflict to doctrine/dbal < 3.6 for loupe adapter

alexander-schranz:
When upgrading to Loupe 0.4 when is released this conflict can be resolved as already be handled by @loupe-php itself. /cc @Toflar


alexander-schranz:
This can be removed when @loupe-php 0.4 is released and adapter updated to it:  https://github.com/loupe-php/loupe/pull/43 is released

PR #270, opened 2023-10-27
Fix code example for SearchCondition

MDevster:


alexander-schranz:
Looks like there is currently an error with the redis extension that be ignored. Thx for the Typo fix.

Issue #276, opened 2023-11-11
[Meilisearch] Incorrect conversion of Boolean values


manuelderuiter:
First of all, thank you for creating this wonderful library! We tried to create an implementation with Meilisearch as it's included in Devenv and I have to say so far we are impressed. 👍

I noticed that there are some todo's left in the MeilisearchSearcher and I'm hoping you're open for a PR since I was able to fix this issue by escaping the values in the ``search()`` method of ``MeilisearchSearcher``.

**Steps to reproduce:**
1. Create an index:
```php
use Schranz\Search\SEAL\Schema\Field;
use Schranz\Search\SEAL\Schema\Index;

new Index('article', [
    'id' => new Field\IdentifierField('id'),
    'is_online' => new Field\BooleanField('is_online', filterable: true),
]);
```

2. Add some items.
3. Create a query builder:
```php
use Schranz\Search\SEAL\EngineRegistry;
use Schranz\Search\SEAL\Search\Condition;
use Schranz\Search\SEAL\Search\SearchBuilder;

$qb = $this->engineRegistry->getEngine('poc')->createSearchBuilder()->addIndex('article')
->addFilter(new Condition\EqualCondition('is_online', $isOnline))
```

**Current outcome**
```php
// seal-meilisearch-adapter/src/MeilisearchSearcher.php line 104
$data = $searchIndex->search($query, $searchParams)->toArray();
$searchParams = [
    'q' => '',
    'filter' => 'is_online = 1'
];
```

**Expected outcome**
```php
// seal-meilisearch-adapter/src/MeilisearchSearcher.php line 104
$data = $searchIndex->search($query, $searchParams)->toArray();
$searchParams = [
    'q' => '',
    'filter' => 'is_online = true'
];
```

alexander-schranz:
@manuelderuiter first thank you for your feedback and thank you for reporting your issues.

Looks like we need to handle the Boolean field differently then a integer field for Meilisearch. I shortly added a test case in #277. If you want to contribute the fix for Meilisaerch I'm happy if you open a pull request for it, just let me know if you want to tackle it?

I can have a look at the other search engines look like the issue is not only Meilisearch specific and we need to handle a Boolean Filter specially also for other engines.

[...]

PR #277, opened 2023-11-12
Add new test case for equal boolean field

alexander-schranz:
## Why?

See #276


alexander-schranz:
Merging this into an own branch [bugfix/boolean-filter](https://github.com/schranz-search/schranz-search/tree/bugfix/boolean-filter) which will be used as target branch for collect all Pull Requests for fixing the boolean filters.

PR #279, opened 2023-11-12
Fix BooleanCondition for different Adapters

alexander-schranz:
# This Pull Request was replaced by #284


alexander-schranz:
Reverted accidently merge, replaced by #284

Issue #292, opened 2023-11-13
[Meilisearch] [Algolia] [Loupe] [Typsense] [RediSearch] [Solr] Escaping of Text based filters


alexander-schranz:
We should escape the input values correctly for different search engines:

 - [x] Loupe: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-loupe-adapter/src/LoupeSearcher.php#L80
 - [x] Meilisearch: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-meilisearch-adapter/src/MeilisearchSearcher.php#L130
 - [x] Algolia: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-algolia-adapter/src/AlgoliaSearcher.php#L143
 - [x] Typesense: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-typesense-adapter/src/TypesenseSearcher.php#L134
 - [x] Solr: Already escaped (still found some issues): https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-solr-adapter/src/SolrSearcher.php#L85
 - [x] Redisearch: Already escaped (still found some issues): https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-redisearch-adapter/src/RediSearchSearcher.php#L204
 - [x] Elasticsearch: Json based no escaping required: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php#L85
 - [x] Opensearch: Json based no escaping required: https://github.com/schranz-search/schranz-search/blob/68c24aa338bda51786a693e2e4fc7e067228247d/packages/seal-opensearch-adapter/src/OpensearchSearcher.php#L75

alexander-schranz:
@manuelderuiter FYI this is now also fixed.

alexander-schranz:
There still one issue with `Solr` when `\r` exists: https://github.com/schranz-search/schranz-search/pull/302

alexander-schranz:
The Solr issue is now also solved in #302

PR #296, opened 2023-11-16
Fix escaping of string based filters in LoupeSearcher

alexander-schranz:
see #292 and #293

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeSearcher.php
+++ b/packages/seal-loupe-adapter/src/LoupeSearcher.php
@@ -75,14 +75,14 @@ final class LoupeSearcher implements SearcherInterface
         $filters = [];
         foreach ($search->filters as $filter) {
             match (true) {
-                $filter instanceof Condition\IdentifierCondition => $filters[] = $index->getIdentifierField()->name . ' = \'' . $filter->identifier . '\'', // TODO escape?
+                $filter instanceof Condition\IdentifierCondition => $filters[] = $index->getIdentifierField()->name . ' = ' . $this->escapeFilterValue($filter->identifier),
                 $filter instanceof Condition\SearchCondition => $query = $filter->query,
-                $filter instanceof Condition\EqualCondition => $filters[] = $filter->field . ' = \'' . $filter->value . '\'', // TODO escape?
-                $filter instanceof Condition\NotEqualCondition => $filters[] = $filter->field . ' != \'' . $filter->value . '\'', // TODO escape?
-                $filter instanceof Condition\GreaterThanCondition => $filters[] = $filter->field . ' > ' . $filter->value, // TODO escape?
-                $filter instanceof Condition\GreaterThanEqualCondition => $filters[] = $filter->field . ' >= ' . $filter->value, // TODO escape?
-                $filter instanceof Condition\LessThanCondition => $filters[] = $filter->field . ' < ' . $filter->value, // TODO escape?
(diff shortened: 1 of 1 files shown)


packages/seal-loupe-adapter/src/LoupeSearcher.php
```diff
@@ -122,6 +122,15 @@ public function search(Search $search): Result
         );
     }
 
+    private function escapeFilterValue(string|int|float|bool $value): string
+    {
+        return match (true) {
+            \is_string($value) => '"' . \addslashes($value) . '"',
```
alexander-schranz: `addslashes` sadly not enough still run into:

```
Loupe\Loupe\Exception\FilterFormatException: Col 7: Error: Expected valid string or float value, got '"'
```

The test uses a very special string:

```php
$value = "The 17\" O'Conner && O`Series \n OR a || 1%2 book?";
```

@Toflar do you have a hint for me? PS: Loupe 0.4

```bash
object(Loupe\Loupe\SearchParameters)#524 (10) {
  ["attributesToHighlight":"Loupe\Loupe\SearchParameters":private]=>
  array(0) {
  }
  ["attributesToRetrieve":"Loupe\Loupe\SearchParameters":private]=>
  array(1) {
    [0]=>
    string(1) "*"
  }
  ["attributesToSearchOn":"Loupe\Loupe\SearchParameters":private]=>
  array(1) {
    [0]=>
    string(1) "*"
  }
  ["filter":"Loupe\Loupe\SearchParameters":private]=>
  string(59) "tags = "The 17\" O\'Conner && O`Series
 OR a || 1%2 book?""
  ["hitsPerPage":"Loupe\Loupe\SearchParameters":private]=>
  int(20)
  ["page":"Loupe\Loupe\SearchParameters":private]=>
  int(1)
  ["query":"Loupe\Loupe\SearchParameters":private]=>
  string(0) ""
  ["showMatchesPosition":"Loupe\Loupe\SearchParameters":private]=>
  bool(false)
  ["showRankingScore":"Loupe\Loupe\SearchParameters":private]=>
  bool(false)
  ["sort":"Loupe\Loupe\SearchParameters":private]=>
  array(1) {
    [0]=>
    string(15) "_relevance:desc"
  }
}
```

Toflar: I think Loupe should provide escaping itself for the filter. I'll check if I can implement this.

alexander-schranz: If I use '"'  around the string and still escaping the `"` in `17"` to `17\"` does not have effect and fails with:

>  Loupe\Loupe\Exception\FilterFormatException: Col 7: Error: Expected valid string or float value, got '"'

If using `'` it fails in `'O'Conner'` even it is escaped to `'O\'Conner'`:

> Loupe\Loupe\Exception\FilterFormatException: Col 20: Error: Expected filterable attribute, got 'Conner'

@toflar a provided escaping function would be great.

Toflar: Yes, the parser is not able to do this atm.

alexander-schranz: I see, keep me uptodate :)

Toflar: Can you test with https://github.com/loupe-php/loupe/pull/54? :)

PR #302, opened 2023-11-20
Fix query filtering of "\r" containing values in Solr

alexander-schranz:
Issues:

 - [x] Solr: https://github.com/solariumphp/solarium/issues/1104


alexander-schranz:
Thx to @thomascorthals we could fix this issue via replacing `\r\n` and `\r` with `\n` this is working because we are in control how the data get indexed.

For more information see: https://github.com/solariumphp/solarium/issues/1104#issuecomment-1821719852

Issue #313, opened 2023-12-12
[Elasticsearch] Combine SearchCondition and EqualCondition resulting in erroneous ELS request


dhirtzbruch:
I am not sure if this might be an Elastic version or some other issue. We are currently using ELS 7.0, but I setup a blank ELS 8.0 too, which resultet in the same issue.

The following code works perfect:

```php
        $raw = $this->engine->createSearchBuilder()
                            ->addIndex(self::INDEX_NAME)
                            ->addFilter(new SearchCondition($search))
                            ->limit($limit)
                            ->offset($offset)
                            ->getResult();
```

Resulting JSON body:

```json
{
    sort: [ ],
    query: {
        query_string: {
            query: "TES*"
        }
    }
}
```

Changing the search configuration to the following code also works:

```php
        $raw = $this->engine->createSearchBuilder()
                            ->addIndex(self::INDEX_NAME)
                            ->addFilter(new EqualCondition('branch', $branch->getShortname()))
                            ->getResult();
```

Resulting JSON body:

```json
{
    sort: [ ],
    query: {
        bool: {
            filter: [
                {
                    term: {
                        branch.raw: {
                        value: "DE"
                        }
                    }
                }
            ]
        }
    }
}
```

However, combining both:

```php
        $raw = $this->engine->createSearchBuilder()
                            ->addIndex(self::INDEX_NAME)
                            ->addFilter(new SearchCondition($search))
                            ->addFilter(new EqualCondition('branch', $branch->getShortname()))
                            ->getResult();
```

results in the following JSON body:

```json
{
    sort: [ ],
        query: {
        query_string: {
            query: "TES*"
        },
        bool: {
            filter: [
                {
                    term: {
                        branch.raw: {
                            value: "DE"
                        }
                    }
                }
            ]
        }
    }
}
```

which at least both available ELS instances (7.0 and 8.0) don't seem to like...

```json
{"error":{"root_cause":[{"type":"parsing_exception","reason":"[query_string] malformed query, expected [END_OBJECT] but found [FIELD_NAME]","line":1,"col":53}],"type":"parsing_exception","reason":"[query_string] malformed query, expected [END_OBJECT] but found [FIELD_NAME]","line":1,"col":53},"status":400}
```

As far as I can see, all three variants work when changing the code in `ElasticsearchSearcher`:

```php
// old
//                $filter instanceof Condition\SearchCondition => $query['query_string']['query'] = $filter->query,
// new
[...]

PR #314, opened 2023-12-12
Fix usage of SearchCondition with other Conditions in Elasticsearch and Opensearch

dhirtzbruch:
Should IMHO fix issue #313


alexander-schranz:
Looks like the same issue appears on `Opensearch` can you implement the same mechanism on that Adapter? The `Lint` should be fixable via `php-cs-fixer` you can run it via `composer fix` command.

dhirtzbruch:
@alexander-schranz Any idea why the opensearch test keeps failing in Github action? It works on my machine... ;-)

alexander-schranz:
> Any idea why the opensearch test keeps failing in Github action? It works on my machine... ;-)

It sadly a race condition [problem of Github CI](https://github.com/actions/checkout/issues/518) not cloning the correct commit, you can ignore it as long as locally the issues are fixed.

dhirtzbruch:
Oh, thanks - then it's working from my point of view.

alexander-schranz:
@dhirtzbruch Thank you for the test and fix, good work 👍

Issue #320, opened 2023-12-15
[Core] Parameter `index_name_prefix` ignored when using one schema file per index


dhirtzbruch:
### Discussed in https://github.com/schranz-search/schranz-search/discussions/319

<div type='discussions-op-text'>

<sup>Originally posted by **dhirtzbruch** December 15, 2023</sup>
Can someone tell me what the `index_name_prefix` in the SymfonyBundle is used for? By the parameter name I thought it should be possible to add a prefix to any index used when creating indexes as well as when querying the data - but I could not get that to work.

Use case would be using the same ElasticSearch cluster for staging / prod environments, and just automatically having the prefix added through the bundle configuration.</div>

<div type='discussions-op-text'>

<sup>Answer by **alexander-schranz**</sup>
It was created for that usecase, specially running tests in another env with a prefix. But if I look there is currently the prefix only used when there are multiple files. So think there is a bug here:
</div>

dhirtzbruch:
Should be all done

alexander-schranz:
fixed via #321

PR #321, opened 2023-12-15
Parameter index_name_prefix ignored when using one schema file per index

dhirtzbruch:
Fixes issue #320


dhirtzbruch:
Should be all done

alexander-schranz:
@dhirtzbruch thank you!

PR #323, opened 2023-12-22
Add PHP 8.3 to test matrix

ker0x:


alexander-schranz:
To keep the CI small I will not run PHP 8.2 in the CI in favor of 8.3 there are mosly no need as we run still 8.1 on lowest.

alexander-schranz:
@ker0x Thank you!

PR #335, opened 2024-01-09
Add Support for Loupe 0.5

### LINKED ISSUE #268: [Loupe] please add support for version 0.5.0

alexander-schranz:
As stated in the CHANGELOG of Loupe 0.5, Loupe requires now a directory instead of a DB: https://github.com/loupe-php/loupe/blob/main/CHANGELOG.md

So instead of `var/indexes/<index_name>.db`, we have now `var/indexes/<index_name>/loupe.db`.

Beside our own configuration which we store in the directory as `config.loupe`, Loupe itself stores other files like `loupe.db-shm`, `loupe.db-wal`, `state_set.php`, in that directory.

For dropping a index we need first remove all files inside the `var/indexes/<index_name>/` directory and then remove the empty dir via rmdir.

Support for previous versions of Loupe is dropped.

fixes #268

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeHelper.php
+++ b/packages/seal-loupe-adapter/src/LoupeHelper.php
@@ -51,21 +51,34 @@ final class LoupeHelper
 
     public function existIndex(Index $index): bool
     {
-        return \file_exists($this->getIndexFile($index));
+        $indexDirectory = $this->getIndexDirectory($index);
+
+        return \file_exists($indexDirectory);
     }
 
     public function dropIndex(Index $index): void
     {
         if ($this->existIndex($index)) {
-            \unlink($this->getIndexFile($index));
-            \unlink($this->getConfigurationFile($index));
+            $indexDirectory = $this->getIndexDirectory($index);
+
+            // beside the .db and our own .loupe file there exists other files which we need to remove
+            $iterator = new \FilesystemIterator($indexDirectory);
+            foreach ($iterator as $fileInfo) {
+                $filePath = (string) $fileInfo;
+                if (\is_file($filePath)) {
+                    \unlink($filePath);
+                }
+            }
+
+            \rmdir($indexDirectory);
         }
     }
 
     public function createIndex(Index $index): void
     {
-        if (!\file_exists($this->directory)) {
-            \mkdir($this->directory, recursive: true);
+        $indexDirectory = $this->getIndexDirectory($index);
+        if (!\file_exists($indexDirectory)) {
+            \mkdir($indexDirectory, recursive: true);
         }
 
         // dumping the configuration allows us to search and index without knowing the configuration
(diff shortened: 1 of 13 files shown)


alexander-schranz:
@Toflar can you maybe check if the changes using a directory per index make sense for you and the changes I did for the code.

packages/seal-loupe-adapter/src/LoupeHelper.php
```diff
@@ -51,29 +51,41 @@ public function getLoupe(Index $index): Loupe
...
     }
 
     public function dropIndex(Index $index): void
     {
         if ($this->existIndex($index)) {
-            \unlink($this->getIndexFile($index));
-            \unlink($this->getConfigurationFile($index));
+            $indexDirectory = $this->getIndexDirectory($index);
+
+            // beside the .db and our own .loupe file there exists other files which we need to remove
```
Toflar, line 64: Why this complicated? Do you also need a directory to store configuration? I would recommend keeping the loupe directory and your own config directory separate (or have the loupe one inside yours).

alexander-schranz: Changing the configuration (searchable fields, sortable fields, ...) is in `SEAL` a active manual process via a CLI Command. To achieve this in Loupe I'm dump the `Configuration` of Loupe into a File and that is used always, and so the configuration only get updated / index migrated to the new fields after the Command is run.

https://github.com/schranz-search/schranz-search/blob/37da7cc77928b56d7c047be20ded10e4dc7bd3b4/packages/seal-loupe-adapter/src/LoupeHelper.php#L75

So it is just one additonal file beside your existing files, which I called now `config.loupe`.

alexander-schranz: I didn't find another way here to remove a Index. As there is no buildin function in PHP removing a directory containing files. So I think the dropIndex will be still the same, when I introduce additonal directory I just need then do it also recursive.

Toflar: Okay, yes then it looks good. `symfony/filesystem` could do that but I understand you probably want to keep the deps low.

alexander-schranz: Before the structure was for two index movies, news:

 - var/indexes/movies.db
 - var/indexes/movies.loupe
 - var/indexes/news.db
 - var/indexes/news.loupe
 
 Now it is:
 
 - var/indexes/movies/loupe.db
 - var/indexes/movies/... # other loupe files
 - var/indexes/movies/config.loupe
 - var/indexes/news/loupe.db
 - var/indexes/news/... # other loupe files
 - var/indexes/news/config.loupe

alexander-schranz: Thx, yeah want currently avoid additional deps for such things :)

PR #372, opened 2024-03-05
Fix Format fields with underscore in Loupe Adapter

### LINKED ISSUE #368: [Loupe] Object fields are not normalized with underscore in sort conditions
With the following schema:

```php
return new Index('guest_session', [
    'uuid' => new Field\IdentifierField('uuid'),
    'dates' => new Field\ObjectField('dates', [
        'start_timestamp' => new Field\IntegerField('start_timestamp', sortable: true),
        'end_timestamp' => new Field\IntegerField('end_timestamp', sortable: true),
    ]),
]);
```

`dates.start_timestamp` is normalized to `dates_start_timestamp` when the index is created (https://github.com/schranz-search/schranz-search/blob/0.3/packages/seal-loupe-adapter/src/LoupeHelper.php#L126).

When I apply a sort condition with the `SearchBuilder`, the field is not normalized with an underscore, which throw an exception.

```php
use Schranz\Search\SEAL\Search\Condition;

$result = $this->engine->createSearchBuilder()
    ->addIndex('guest_session')
    ->addSortBy('dates.start_timestamp', 'asc')
    ->getResult();
```
```
Loupe\Loupe\Exception\SortFormatException: Cannot sort by "dates.start_timestamp". It must be defined as sortable attribute.
```

ker0x:
Fix #368

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeHelper.php
+++ b/packages/seal-loupe-adapter/src/LoupeHelper.php
@@ -23,7 +23,7 @@ use Schranz\Search\SEAL\Schema\Index;
  */
 final class LoupeHelper
 {
-    public const SEPERATOR = '_';
+    public const SEPARATOR = '_';
     public const SOURCE_FIELD = 'l_source'; // fields are not allowed to begin with `_`
 
     /**
@@ -98,6 +98,14 @@ final class LoupeHelper
         $this->loupe = [];
     }
 
+    /**
+     * @internal
+     */
+    public function formatField(string $field): string
+    {
+        return \str_replace('.', self::SEPARATOR, $field);
+    }
+
     private function createLoupe(Index $index, Configuration|null $configuration = null): Loupe
     {
         if (!$configuration instanceof Configuration) {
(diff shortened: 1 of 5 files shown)


packages/seal-loupe-adapter/src/LoupeHelper.php
```diff
@@ -98,6 +98,11 @@ public function reset(): void
         $this->loupe = [];
     }
 
+    public function formatField(string $field): string
```
alexander-schranz, line 104: We should mark this method as `@internal`

```suggestion
    /**
     * @internal
     */
    public function formatField(string $field): string
```

integrations/spiral/src/Console/IndexCreateCommand.php
```diff
@@ -29,10 +29,10 @@
 final class IndexCreateCommand extends Command
 {
     #[Option(name: 'engine', mode: InputOption::VALUE_REQUIRED, description: 'The name of the engine')]
-    private string|null $engineName = null;
+    private ?string $engineName = null;
```
alexander-schranz: The code style was updated. Please rebase to 0.3

alexander-schranz (changes requested):
Thx for the pull request.

Looks good, just some small change request to get this merged.

I need to check if we can maybe with the current test fixtures create a reproducable failing test case. But I think at current state the `memory` adapter is very minimal doesn't supported already this kind of nested properties.

ker0x:
@alexander-schranz i'va rebase and apply your suggestions 😉

alexander-schranz:
@ker0x thank you!

PR #405, opened 2024-07-01
Fix yii dependencies in example

alexander-schranz:
Looks like the `yiisoft/yii-debug` notlonger has branch alias or branch for `3.0` / `3.x` so composer can not resolve it longer. /cc @samdark


samdark:
Yes. We're getting rid of aliases closer to release.

PR #427, opened 2024-10-06
Add stricter validation for field names

### LINKED ISSUE #426: Identifiers for meilisearch and probably others need to be checked
If you use a document with the ID `foo.bar`, it will cause issues because meilisearch does not support `.` in identifiers: https://www.meilisearch.com/docs/learn/getting_started/primary_key#formatting-the-document-id

However, other search engines do. For compatibility, I guess the marshaller would need to normalize these identifiers.
What do you think?

alexander-schranz:
This outputs a error message for an unexpected input for the field name. We validate currently two things.

 - field Key === field name (this may be change in future)
 - field name only supports a-z, A-Z, 0-9 and _ (see #426)

fix #426 

FYI @Toflar

### DIFF
--- a/.github/workflows/callable-test.yml
+++ b/.github/workflows/callable-test.yml
@@ -44,7 +44,7 @@ jobs:
               with:
                   php-version: ${{ matrix.php-version }}
                   tools: 'composer:v2'
-                  ini-values: memory_limit=-1
+                  ini-values: memory_limit=-1, zend.assertions=1
                   coverage: none
 
             - name: Install composer dependencies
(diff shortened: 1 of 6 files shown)


alexander-schranz:
PS: I'm using internal `assert` of PHP to avoid overheads in validation for production. For CI using the widely used https://github.com/shivammathur/setup-php It is required to configure `zend.assertions=1` via:

```yaml
            - name: Install and configure PHP
              uses: shivammathur/setup-php@v2
              with:
                  php-version: ${{ matrix.php-version }}
                  tools: 'composer:v2'
                  ini-values: memory_limit=-1, zend.assertions=1
```

FYI: @Toflar

Issue #317, opened 2023-12-13
[Core] Add support for Nested Conditions Filters via AndCondition and OrCondition


alexander-schranz:
It would be great if we could also possibility to have complex filters which are combined via `OrCondition` and also have there again nested `AndCondition` inside.

The condition signature could looke like this:

```php
class OrCondition {
    public readonly array $conditions;
    
    public function __construct(...$conditions)
    {
        $this->conditions = $conditions;
    }
}

class AndCondition {
    public readonly array $conditions;
    
    public function __construct(...$conditions)
    {
        $this->conditions = $conditions;
    }
}
```

The queries in the specific `...Searcher` classes will get more complex. We would require to add some `Recursive functionality` and before implement such feature we require a research if this can even be achieved by all currently supported search engines.

ToshY:
> ...research if this can even be achieved by all currently supported search engines

Done a quick search:

* algolia - [yes](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/combining-boolean-operators/)
* elasticsearch - ?
* opensearch - ?
* loupe - [yes](https://github.com/loupe-php/loupe/blob/main/docs/searching.md#filter)
* meilisearch - [yes](https://www.meilisearch.com/docs/learn/fine_tuning_results/filtering#filter-expressions)
* redisearch - [yes](https://redis.io/docs/interact/search-and-query/query/combined/)
* solr - [yes](https://solr.apache.org/guide/6_6/the-standard-query-parser.html#TheStandardQueryParser-BooleanOperatorsSupportedbytheStandardQueryParser)
* typesense - ?

I'm not sure about elasticsearch, opensearch and typesense.

alexander-schranz:
@ToshY thx for the research:

The others seems also support it:

 - elasticsearch/opensearch can use the Boolquery: https://www.elastic.co/guide/en/elasticsearch/reference/8.11/query-dsl-bool-query.html#score-bool-filter
    - AND -> `bool:must`
    - OR -> `bool:should`
    - Nesting -> `nested bool queries`
 - typesense: https://github.com/typesense/typesense/issues/281
    - AND -> `&&`
    - OR -> `||`
    - Nesting -> `()`

alexander-schranz:
implemented in #317

ToshY:
@alexander-schranz Guess you meant implemented in #433 🙂 

Cool to see this implemented 👍 Will the docs also be updated with some examples on how to use these new conditions?

[...]

PR #433, opened 2024-10-07
Add support for Nested Conditions via AndCondition and OrCondition 

Toflar:
Started implementing https://github.com/schranz-search/schranz-search/issues/317.

Works for Meilisearch and Loupe, the others I haven't got any local setup for at the moment 🙈 
But maybe this can be a helpful start to push it over the finishing line.

---

This adds 2 new conditions.

**AndCondition**

```php
$searchBuilder->addFilter(new Condition\AndCondition($conditionA, $conditionB);
```

**OrCondition**

```php
$searchBuilder->addFilter(new Condition\OrCondition($conditionA, $conditionB);
```

----

**Nested conditions** are now possible:

```php
$searchBuilder->addFilter(
    new Condition\AndCondition(
        new Condition\EqualCondition('tags', 'Tech'),
        new Condition\OrCondition(
            new Condition\EqualCondition('tags', 'UX'),
            new Condition\EqualCondition('isSpecial', false),
        ),
    ),
);
```

### DIFF
--- a/packages/seal/src/Search/Condition/AndCondition.php
+++ b/packages/seal/src/Search/Condition/AndCondition.php
@@ -1,13 +1,18 @@
 <?php
 
+declare(strict_types=1);
+
+/*
+ * This file is part of the Schranz Search package.
+ *
+ * (c) Alexander Schranz <alexander@sulu.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 namespace Schranz\Search\SEAL\Search\Condition;
 
-class AndCondition
+class AndCondition extends AbstractGroupCondition
 {
-    /**
-     * @var array<object>
-     */
-    public function __construct(public readonly array $conditions)
-    {
-    }
 }
--- a/packages/seal/src/Search/Condition/AndCondition.php
+++ b/packages/seal/src/Search/Condition/AndCondition.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Schranz\Search\SEAL\Search\Condition;
+
+class AndCondition
+{
+    /**
+     * @var array<object>
+     */
+    public function __construct(public readonly array $conditions)
+    {
+    }
+}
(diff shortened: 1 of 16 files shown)


alexander-schranz:
@toflar Thank you!

> Works for Meilisearch and Loupe, the others I haven't got any local setup for at the moment 🙈

If you want to give it a try a `docker-compose.yml` exists in every adapters directory if you want to start the service `docker compose up` is enough. What may very similar to Meilisearch is Typesense Adapter: https://github.com/schranz-search/schranz-search/blob/4267ea93ec459e15df084cd1727457dcbd1a4457/packages/seal-typesense-adapter/src/TypesenseSearcher.php#L77-L114

The only problem if I see currently a Geo based Conditions can not be part of a Nested Conditions. But think that is some edgecase we currently can ignore.

packages/seal/src/Search/Condition/AndCondition.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Schranz\Search\SEAL\Search\Condition;
+
+class AndCondition
+{
+    /**
+     * @var array<object>
+     */
+    public function __construct(public readonly array $conditions)
```
alexander-schranz: What you think is more developer friendly go here with an `array` or variable parameters length e.g.:

```php
public function __construct(...$conditions)
```

```php
$condition = new Condition\OrCondition(
    new Condition\AndCondition(
        new Condition\EqualCondition('tags', 'Tech'),
        new Condition\EqualCondition('isSpecial', true),
    ),
    new Condition\AndCondition(
        new Condition\EqualCondition('tags', 'UX'),
        new Condition\EqualCondition('isSpecial', false),
    ),
);
```

Toflar:
I have updated the conditions so that they are easier to use and they also only accept the supported nested conditions. Typesense is implemented as well now.

For the `MemoryAdapter` - am I right that this is just for testing purposes? We could just use Loupe with the in-memory configuration - it's been designed for exactly that and then we wouldn't have to develop all that logic with nested comparisons etc.

alexander-schranz:
I will take care of the memory adapter. At current state I want to keep the memory adapter as it good to have one which don't require any additional dependencies.

Toflar:
Elasticsearch, Opensearch and Algolia are now also supported. I had to adjust the example because Algolia does not support filtering for ORs of ANDs (https://github.com/algolia/algoliasearch-client-php/issues/385).

alexander-schranz:
@Toflar Good work, thank you 👍

PR #435, opened 2024-10-08
Move bulk method from own Interface to IndexerInterface

alexander-schranz:
This moves the bulk method from own interface to indexer interface. So without any checks we can use bulk everywhere as supported by all adapters currently.

None BC Break as old BulkableIndexerInterface was not part of any release yet.


alexander-schranz:
@tacman FYI the BulkableIndexerInterface got removed as not required as we could support bulk operations for all adapters we could safely add it to the `IndexerInterface`.

PR #437, opened 2024-10-08
Fix yii integration tests

alexander-schranz:
Downgrade yii-runner-http as in `yiisoft/app` branch. Else the server factory fails with:

```
  [Yiisoft\Di\NotFoundException] No definition or class found or resolvable for "Yiisoft\Yii\Runner\Http\ServerRequestFactory" while building "Yiisoft\Yii\Runner\Http\ServerRequestFactory".  
```

Also changed some code to use PSR Server Factory now.

### DIFF
--- a/.examples/yii/composer.json
+++ b/.examples/yii/composer.json
@@ -89,7 +89,7 @@
         "yiisoft/yii-http": "^1.0",
         "yiisoft/yii-middleware": "^1.0",
         "yiisoft/yii-runner-console": "^2.0",
-        "yiisoft/yii-runner-http": "^2.0 || ^3.0",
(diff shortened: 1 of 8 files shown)


.examples/yii/composer.json
```diff
@@ -89,7 +89,7 @@
         "yiisoft/yii-http": "^1.0",
         "yiisoft/yii-middleware": "^1.0",
         "yiisoft/yii-runner-console": "^2.0",
-        "yiisoft/yii-runner-http": "^2.0 || ^3.0",
+        "yiisoft/yii-runner-http": "^2.0",
```
alexander-schranz, line 92: @samdark maybe you have a hint how we can keep compatibility here?

PR #438, opened 2024-10-10
Make And and OrConditions readonly

alexander-schranz:
At current state I will keep the AndCondition and OrCondition like all existing conditions readonly to be consistent.

### DIFF
--- a/packages/seal/src/Search/Condition/AbstractGroupCondition.php
+++ b/packages/seal/src/Search/Condition/AbstractGroupCondition.php
@@ -21,7 +21,7 @@ abstract class AbstractGroupCondition
     /**
      * @var array<EqualCondition|GreaterThanCondition|GreaterThanEqualCondition|IdentifierCondition|LessThanCondition|LessThanEqualCondition|NotEqualCondition|AndCondition|OrCondition>
      */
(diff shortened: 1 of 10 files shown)


alexander-schranz:
@Toflar FYI

packages/seal/src/Search/Condition/AbstractGroupCondition.php
```diff
@@ -13,35 +13,21 @@
...
+/**
+ * @internal this class is internal please use AndCondition or OrCondition directly
+ */
 abstract class AbstractGroupCondition
 {
     /**
      * @var array<EqualCondition|GreaterThanCondition|GreaterThanEqualCondition|IdentifierCondition|LessThanCondition|LessThanEqualCondition|NotEqualCondition|AndCondition|OrCondition>
      */
-    private array $conditions = [];
+    public array $conditions = [];
```
Toflar: This will allow to set conditions directly that are not validated against the allowed types. That's why I did not do this.

alexander-schranz: Good catch `readonly` missing :)

PR #441, opened 2024-10-10
Add documentation about AndCondtion and OrCondition filters

### LINKED ISSUE #317: [Core] Add support for Nested Conditions Filters via AndCondition and OrCondition

alexander-schranz:
This add missing docs about the newly added `AndCondition` and `OrCondition` from https://github.com/schranz-search/schranz-search/pull/433 .

fix #317

### DIFF
--- a/docs/search-and-filters/index.rst
+++ b/docs/search-and-filters/index.rst
@@ -228,6 +228,53 @@ The field is required to be marked as ``filterable`` in the index configuration.
     The ``GeoBoundingBoxCondition`` is currently not supported by ``Redisearch`` adapter.
     See `this Github Issue <https://github.com/RediSearch/RediSearch/issues/680>`__ for more information.
 
+OrCondition
+~~~~~~~~~~~
+
+The ``OrCondition`` is used to filter by two or more conditions where at least one condition need to match.
+
+.. code-block:: php
+
+    <?php
+
+    use Schranz\Search\SEAL\Search\Condition;
+
+    $result = $this->engine->createSearchBuilder()
+        ->addIndex('blog')
+        ->addFilter(new Condition\OrCondition(
+            new Condition\GreaterThanCondition('rating', 2.5),
+            new Condition\EqualCondition('isSpecial', true),
+        ))
+        ->getResult();
+
+The fields are required to be marked as ``filterable`` in the index configuration.
+
+AndCondition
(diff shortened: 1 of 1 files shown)


docs/search-and-filters/index.rst
```diff
@@ -228,6 +228,53 @@ The field is required to be marked as ``filterable`` in the index configuration.
...
+        ))
+        ->getResult();
+
+The fields are required to be marked as ``filterable`` in the index configuration.
+
+AndCondition
+~~~~~~~~~~~~
+
+The ``AndCondition`` is used to combine two or multiple conditions where all conditions need to match.
+As by default all conditions are ``AND``` connected it only make sense to use ``AndCondition`` in
```
ToshY, line 256: Think the textual explanation can be improved a bit:

1. There's a trailing backtick too much in ``AND```.
2. Think grammatically the second sentence is a bit off: 
   1. I'm reading the text "*As by default all conditions are ``AND``` connected it only make sense...*" as one single sentence, even though I think there's supposed to be a comma here (after "connected").
   2. Think "*...it only make**s** sense*" is correct (the additional **s**).
   3. Think you can omit "an" (singular) in "...*with an ``OrCondition`` filters.*". Thus "...*with ``OrCondition`` filters.*" seems better.

Overall, I would rephrase the second sentence more like this:

> By default, all conditions are connected with ``AND``, so it only makes sense to use an ``AndCondition`` in combination with ``OrCondition`` filters.

---

It seems very (very) nitpicky to be honest, and while I'm willing to create a PR for it, I'm not sure if it's of any worth. Maybe next time the docs are updated this can be updated as well. It's totally up to you 🙂

Overall the examples are clear and that's what matters.

PR #448, opened 2024-10-15
Add `NotInCondition` Filter

### LINKED ISSUE #445: [Core] Add support for `NotInCondition`
**Suggestion**

The ability to check for multiple values using a `NOT IN` condition would be nice to have. The idea would be roughly be identical to the `NotEqualCondition` except the second argument allows an array (multiple values).

```php
<?php

use Schranz\Search\SEAL\Search\Condition;

$result = $this->engine->createSearchBuilder()
    ->addIndex('blog')
    ->addFilter(new Condition\NotInCondition('tags', ['UI','UX','Frontend']))
    ->getResult();
```

---

This probably would require research to check if all engines allow a kind of "IN".

* algolia - [`AND NOT` chaining](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/combining-boolean-operators/#example-scenarios) (unsure)
* elasticsearch - [`terms` with `must_not`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-terms-query.html)
* opensearch - [`terms` with `must_not`](https://opensearch.org/docs/latest/query-dsl/term/terms/)
* loupe - ?
* meilisearch - [`NOT IN`](https://www.meilisearch.com/docs/learn/filtering_and_sorting/filter_expression_reference#in)
* redisearch - [`|` chaining (negated); `-@x:(foo|bar)`](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/query_syntax/#mapping-common-sql-predicates-to-redis-query-engine) (unsure)
* solr - ?
* typesense - ?

ToshY:
This add supports for `NotInCondition` with the already support for `AndCondition` and `NotEqualCondition` it should be possible to implement a `NotInCondition` in all adapters as `NOT IN` is just a shortcut for `NotEqualCondition` on the same field `AND` connected.

So instead of writing:

```php
new Condition\AndCondition(
    new Condition\NotEqualCondition('tags', 'UI'),
    new Condition\NotEqualCondition('tags', 'UX'),
    new Condition\NotEqualCondition('tags', 'Frontend'),
);
```

It should now be possible to write:

```php
new Condition\NotInCondition('tags', ['UI','UX','Frontend'])
```

Fixes #445 

### Todo

- [x]  Core: Add `NotInCondition`
- [x]  Memory
- [x]  Elasticsearch
- [x]  Opensearch
- [x]  Meilisearch
- [x]  Algolia
- [x]  Loupe
- [x]  Redisearch
- [x]  Solr
- [x]  Typesense

### DIFF
--- a/packages/seal-meilisearch-adapter/src/MeilisearchSearcher.php
+++ b/packages/seal-meilisearch-adapter/src/MeilisearchSearcher.php
@@ -126,6 +126,14 @@ final class MeilisearchSearcher implements SearcherInterface
         };
     }
 
+    private function escapeArrayFilterValues(array $value): string
+    {
+        return implode(
+            ', ',
+            array_map([$this, 'escapeFilterValue'], $value)
+        );
+    }
+
     /**
      * @param object[] $conditions
      */
@@ -143,6 +151,7 @@ final class MeilisearchSearcher implements SearcherInterface
                 $filter instanceof Condition\SearchCondition => $query = $filter->query,
                 $filter instanceof Condition\EqualCondition => $filters[] = $filter->field . ' = ' . $this->escapeFilterValue($filter->value),
                 $filter instanceof Condition\NotEqualCondition => $filters[] = $filter->field . ' != ' . $this->escapeFilterValue($filter->value),
+                $filter instanceof Condition\NotInCondition => $filters[] = $filter->field . ' NOT IN [' . $this->escapeArrayFilterValues($filter->values) . ']',
                 $filter instanceof Condition\GreaterThanCondition => $filters[] = $filter->field . ' > ' . $this->escapeFilterValue($filter->value),
                 $filter instanceof Condition\GreaterThanEqualCondition => $filters[] = $filter->field . ' >= ' . $this->escapeFilterValue($filter->value),
                 $filter instanceof Condition\LessThanCondition => $filters[] = $filter->field . ' < ' . $this->escapeFilterValue($filter->value),
--- a/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
+++ b/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
@@ -161,9 +161,11 @@ final class AlgoliaSearcher implements SearcherInterface
         $filters = [];
 
         foreach ($conditions as $filter) {
-            if ($filter instanceof Condition\InCondition) {
-                $filter = $filter->createOrCondition();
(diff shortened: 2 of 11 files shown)


packages/seal-meilisearch-adapter/src/MeilisearchSearcher.php
```diff
@@ -143,6 +151,7 @@ private function recursiveResolveFilterConditions(Index $index, array $condition
                 $filter instanceof Condition\SearchCondition => $query = $filter->query,
                 $filter instanceof Condition\EqualCondition => $filters[] = $filter->field . ' = ' . $this->escapeFilterValue($filter->value),
                 $filter instanceof Condition\NotEqualCondition => $filters[] = $filter->field . ' != ' . $this->escapeFilterValue($filter->value),
+                $filter instanceof Condition\NotInCondition => $filters[] = $filter->field . ' NOT IN [' . $this->escapeArrayFilterValues($filter->values) . ']',
```
ToshY, line 154: Syntax specicially for Meilisearch for `IN` and `NOT IN` according to docs: https://www.meilisearch.com/docs/learn/filtering_and_sorting/filter_expression_reference#in.

Could also use the `$filter->createAndCondition()` instead. Not sure what's preferred, the generic way or the one more closely to the adapter.

---

**Edit**

I now see that the `LoupeSearcher` already does this similiarly with `IN` / `NOT IN` (and also has `array_map` which i separated to a separate method).

https://github.com/alexander-schranz/schranz-search/blob/b8e63a48d340b66bc298d55904d9a79767440330/packages/seal-loupe-adapter/src/LoupeSearcher.php#L153

ToshY:
@alexander-schranz I'm not sure why some tests fail sometimes in CI. I have run the tests locally with docker for the following adapters: `meilisearch`, `elasticsearch`, `memory` and `solr`, and they were ok.

alexander-schranz:
@ToshY there is currently an issue with the checkout process going on the false commit hash. You can retrigger the CI with additional commit push `git commit --allow-empty -m "Update base branch"`

alexander-schranz:
I adopted the test case the order for none "searches" can be differently by the search engines so the testcase just need to check if the expected documents exists not if they are in the same order. This way also the other adapters seems to work fine.

alexander-schranz:
@ToshY all green 🎉 . Thank you for your contribution.

PR #452, opened 2024-10-23
Add Windows support for creating search adapters

zoglo:
Another try but now in the right repo :)

### Description

Absolute paths on windows fail due to an undefined array key within the following line not having a scheme at all;
https://github.com/schranz-search/schranz-search/blob/d91446a972b7bcc894abfd803eff40fc7dd3c09c/packages/seal/src/Adapter/AdapterFactory.php#L140

This PR adds `Windows` support for the path resolution since `$dsn` would fail later due to the second `parse_url` attempt returning `false` again.

In my case, the following `$dsn` would fail because it won't detect any scheme at all:
```loupe://C:\path\project\var\indexes```
The early return is done before adding the fallback where it adds the adapter name with an `@´  :')

Unsure if we should go for the `PHP_OS_FAMILY` check and see if it's windows in this case 🤔?

### DIFF
--- a/packages/seal/src/Adapter/AdapterFactory.php
+++ b/packages/seal/src/Adapter/AdapterFactory.php
@@ -87,7 +87,17 @@ final class AdapterFactory
 
             if (\str_contains($dsn, ':///')) {
                 // make DSN like loupe:///full/path/project/var/indexes parseable
-                $dsn = \str_replace(':///', '://' . $adapterName . '/', $dsn . $query);
+                $dsn = \str_replace(':///', '://' . $adapterName . '/', $dsn. $query);
+            } elseif (\DIRECTORY_SEPARATOR === '\\') {
+                // might be Windows and contain an absolute path like loupe://C:\path\project\var\indexes which will fail when parse_url is used
+                $dsnParts = \explode('://', $dsn);
+
+                return [
+                    'scheme' => $dsnParts[0] ?? '',
+                    'host' => '',
+                    'path' => $dsnParts[1] ?? '',
+                    'query' => [],
+                ];
             } else {
                 $dsn = $dsn . '@' . $adapterName . $query;
             }
--- a/packages/seal/src/Adapter/AdapterFactory.php
+++ b/packages/seal/src/Adapter/AdapterFactory.php
@@ -85,19 +85,14 @@ final class AdapterFactory
                 $query = '?' . $query;
             }
 
+            $isWindowsLikePath = false;
             if (\str_contains($dsn, ':///')) {
                 // make DSN like loupe:///full/path/project/var/indexes parseable
-                $dsn = \str_replace(':///', '://' . $adapterName . '/', $dsn. $query);
-            } elseif (\DIRECTORY_SEPARATOR === '\\') {
-                // might be Windows and contain an absolute path like loupe://C:\path\project\var\indexes which will fail when parse_url is used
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/seal/src/Adapter/AdapterFactory.php [hunk 1]
```diff
@@ -88,6 +88,16 @@ public function parseDsn(string $dsn): array
...
                 $dsn = \str_replace(':///', '://' . $adapterName . '/', $dsn);
+            } elseif (\DIRECTORY_SEPARATOR === '\\') {
+                // might be Windows and contain an absolute path like loupe://C:\path\project\var\indexes which will fail when parse_url is used
+                $dsnParts = \explode('://', $dsn);
+
+                return [
+                    'scheme' => $dsnParts[0] ?? '',
+                    'host' => '',
+                    'path' => $dsnParts[1] ?? '',
+                    'query' => [],
```
alexander-schranz: Is there any options were we can just convert the DSN into a format supported by `parse_url` else we would handle here also query and fragment parameter correctly:

```
loupe:///full/path/project/var/indexes?key=value#fragment
```

zoglo: Maybe we could modify the path to be like this, wdyt:
```php
loupe:/C:\path\project\var\indexes?key=value#fragment

Array
(
    [scheme] => loupe
    [path] => /C:\path\project\var\indexes
    [query] => key=value
    [fragment] => fragment
)
```

That way, it would at least not fail nor consider the `C:\` directory as a host.
We'd only have to remove the leading slash from the path then.

See: https://3v4l.org/kI6B3

```php
<?php

function printOutput($output)
{
    echo (is_array($output) ? print_r($output, true) : 'failing') . "\r\n\n";
}

$unixPath    = 'loupe://full/path/project/var/indexes?key=value#fragment';
$windowsPath = 'loupe://C:\path\project\var\indexes?key=value#fragment';


// Normal unix path

printOutput(parse_url($unixPath));

// this one does not work
printOutput(parse_url($windowsPath));


//however this would work if we modify it later on

$windowsPathModified = 'loupe:/C:\path\project\var\indexes?key=value#fragment';
printOutput(parse_url($windowsPathModified));
```

```
Array
(
    [scheme] => loupe
    [host] => full
    [path] => /path/project/var/indexes
    [query] => key=value
    [fragment] => fragment
)


failing

Array
(
    [scheme] => loupe
    [path] => /C:\path\project\var\indexes
    [query] => key=value
    [fragment] => fragment
)
```

zoglo: See 085fc7efdbbff987a1b65b5340feaa9164b85d48

zoglo:
See 085fc7efdbbff987a1b65b5340feaa9164b85d48
I left the explanation here: https://github.com/schranz-search/schranz-search/pull/452#discussion_r1813306240

alexander-schranz:
@zoglo did refactor it to `rawurlencode` the part and decode it later, as I stumbled over other issues in the later process. I'm not sure about the detection based on DIRECTORY_PATH or OS as may the "build" system for containers may a different as where being hosted. So I did manually detection by check if it starts with `<letter>:\` this also let us test it in the Linux CI. With the new `AdapterFactoryTest` it should now be a lot easier to add more tests cases and avoid regressions here.

Thx for your work on this. Let me know if you stumble over more issues.

PR #460, opened 2024-11-21
Introducing partial reindexes (BC Break)

Toflar:
# BC Breaks

To upgrade add the new ReindexConfig parameter to the your ReindexProvider:

```diff
-    public function provide(): \Generator
+    public function provide(ReindexConfig $reindexConfig): \Generator
```

```diff
+$reindexConfig = ReindexConfig::create()
+    ->withIndex('blog')
+    ->withDrop(true)
+    ->withBulkSize(100);

 $engine->reindex(
     iterable $reindexProviders,
+    $reindexConfig,
-    string|null $index = null,
-    bool $dropIndex = false,
-    int $bulkSize = 100,
     callable|null $progressCallback = null,
 );
```


packages/seal/src/Reindex/PartialReindexConfig.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CmsIg\Seal\Reindex;
+
+final class PartialReindexConfig
+{
+    public function __construct(private \DateTimeInterface|\Closure $dateTimeBoundary, private \Generator|\Closure $identifiers)
```
alexander-schranz: Can you give me some background information about the cases when this are closures or generators?

alexander-schranz (approved):

alexander-schranz:
Still unsure about the ReindexConfig but I think we can currently go with it.

PR #462, opened 2024-12-07
Add prototype for Galach query translator implementation

alexander-schranz:
While visiting the #SymfonyCon I think I had a quick chat with @pspanja who showed me: https://github.com/netgen/query-translator.

Think it would be a nice have a package which supports the generation of Filters out of the Galach query language.

This is just a quick prototype and if somebody want to continue the work let me know.

#symfonyconhackday

### DIFF
--- a/packages/seal-query-translator/composer.json
+++ b/packages/seal-query-translator/composer.json
@@ -1,0 +1,78 @@
+{
+    "name": "cmsig/seal-query-translator",
+    "description": "Add support for string query translating for SEAL SearchBuilder.",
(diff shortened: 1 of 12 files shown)


packages/seal-query-translator/composer.json
```diff
@@ -0,0 +1,78 @@
+{
+    "name": "cmsig/seal-query-translator",
```
alexander-schranz, line 2: Not sure about the name yet it generate / create filters out of a Galach query language using netgen/query-translator package.

PR #465, opened 2024-12-19
Make sure identifiers are empty

Toflar:
Fix and add missing parts of https://github.com/PHP-CMSIG/search/pull/460

### DIFF
--- a/integrations/laravel/src/Console/ReindexCommand.php
+++ b/integrations/laravel/src/Console/ReindexCommand.php
@@ -28,7 +28,7 @@ final class ReindexCommand extends Command
      *
      * @var string
      */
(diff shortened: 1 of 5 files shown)


alexander-schranz:
Looks like we missed to add this to the other `ReindexCommands` also?

Toflar:
I am not familiar with any of the other frameworks so I couldn't provide this.

alexander-schranz:
@Toflar yii, mezzio is directly symfony/console so just a copy of that lines. Spiral, Laravel are symfony/console based so just a small adjustment about how the param is defined 😄

alexander-schranz (approved):

Issue #463, opened 2024-12-09
AutoSetup for Indexes


alexander-schranz:
There may an option to auto setup a index if it does not yet exists.

alexander-schranz:
> From https://github.com/PHP-CMSIG/search/pull/470#issuecomment-2567862902

It seems like we are not possbile to standardize this as a lot of Search Engines do not fail when try to add documents to a none existing Index. While for example Elasticsearch and Opensearch have some settings to force error on none created Indexes. Search engines like Algolia and Meilisearch do not have this options. Same for Redisearch which even allow to add Documents but as no Index is created search will still fail.

Sadly this also means we can not force that a Index is created with the correct settings and so can fail if something isn't correct. As we are not able to standardize a index not found exception we are also not able to implement auto create for all engines correctly. So the search engines will behave differently if you try to add a document to none existing Engine.

As request by @Toflar we will support Autocreate for Indexes of Loupe in https://github.com/PHP-CMSIG/search/pull/471.

---

Autosetup with correct configuration:

 - Loupe (see https://github.com/PHP-CMSIG/search/pull/471)
 - Memory (see https://github.com/PHP-CMSIG/search/pull/471)

Autosetup with default search engine defaults (can end in unexpected search results as searchable, filterable, sortable are not correctly configured in the auto creation of index):

 - Elasticsearch 
 - Opensearch
 - Algolia
 - Meilisearch

Partial support:

 - Redisearch (still not searchable only add, delete)

Non support:

 - Typesense
 - Solr

alexander-schranz:
Added wontfix as we are not able to achieve what we want currently, creating the Indexes with the correct configurations. For this we would require #470 which Meilisearch, Algolia, Redisearch didn't support, Elasticsearch and Opensearch only with special none default settings.

PR #470, opened 2024-12-27
Introduce IndexNotFoundException (BC Break)

alexander-schranz:
This PR should let us check if all search engines throw a not found exception which we may can catch and rethrow our own exception to maybe build up a auto setup in future (see #463).

## BC Breaks

New recommended way to setup `Elasticsearch` and `Opensearch` is to disable auto create of indexes, this avoids that a false Schema of your index is created in any case:

```diff
services:
  elasticsearch:
    image: # ...
    environment:
      # ...
+     action.auto_create_index: 'false'
    # ...
```

Existing clusters can be updated with:

```bash
curl -X PUT "http://localhost:9200/_cluster/settings" \
-H "Content-Type: application/json" \
-d '{
  "persistent": {
    "action.auto_create_index": "false"
  }
}'
```

## Todo

 - [x] ⚪ Memory (custom handeled)
 - [x] ⚪ Loupe (custom handeled / by default autocreated with custom settings)
 - [x] ⚠️ Elasticsearch (only with custom setting we get error or 404 else autocreated with basic engine specific  settings)
 - [x] ⚠️ Opensearch (not tested but think same as Elasticsearch)
 - [ ] ❌ Algolia (no error thrown - autocreated with basic engine specific  settings)
 - [ ] ❌ Meilisearch (no error thrown - autocreated with basic engine specific settings)
 - [ ] ❌ Redisearch (no error thrown - document created but Index not as 2 JSON.SET and FT are two different modules)
 - [ ] ⚪ Solr (throws 404)
 - [ ] ⚪ Typesense (throws a NotFoundObject exception)

### DIFF
--- a/.examples/laravel/docker-compose.yml
+++ b/.examples/laravel/docker-compose.yml
@@ -5,6 +5,7 @@ services:
       discovery.type: single-node
       xpack.security.enabled: 'false'
       cluster.routing.allocation.disk.threshold_enabled: 'false'
+      action.auto_create_index: 'false'
     ports:
       - "9200:9200"
     healthcheck:
@@ -34,6 +35,7 @@ services:
     environment:
       discovery.type: single-node
       cluster.routing.allocation.disk.threshold_enabled: 'false'
+      action.auto_create_index: 'false'
       DISABLE_SECURITY_PLUGIN: true
       http.port: 9201
     ports:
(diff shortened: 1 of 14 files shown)


alexander-schranz:
It seems like we are not possbile to standardize this as a lot of Search Engines do not fail when try to add documents to a none existing Index. While for example Elasticsearch and Opensearch have some settings to force error on none created Indexes. Search engines like Algolia and Meilisearch do not have this options. Same for Redisearch which even allow to add Documents but as no Index is created search will still fail.

Sadly this also means we can not force that a Index is created with the correct settings and so can fail if something isn't correct. As we are not able to standardize a index not found exception we are also not able to implement auto create for all engines correctly. So the search engines will behave differently if you try to add a document to none existing Engine.

As request by @Toflar we will support Autocreate for Indexes of Loupe in https://github.com/PHP-CMSIG/search/pull/471.

PR #471, opened 2025-01-02
Add Support for autocreate of an Index for Loupe and Memory Adapter

alexander-schranz:
Sadly it seems like we are not able to standardize IndexNotFoundException (https://github.com/PHP-CMSIG/search/pull/470) for save, delete, bulk or search actions. 

We will still support in Loupe Autocreate of the Index if it does not yet exists.

While Indexes are in some cases autocreated we recommend to create Indexes via the frameworks CLI commands / scripts. Else Search Engine will create there own setting for the indexes and things like searchable, filterable, sortable, does may not work like expected. Why this is not the case for Loupe it is case for other Engine which also support auto create:

 - Elasticsearch
 - Opensearch
 - Algolia
 - Meilisearch
 
They all create indexes with some kind of default settings not the defined mapping of yours.

The following search engines will just fail if you try to save, delete, ... a document from it:

 - Solr
 - Typesense
 
The following search engines will not fail when adding, deleting a document but still no Index exist so nothing is searchable.

 - Redisearch

So fixes only parts of https://github.com/PHP-CMSIG/search/issues/463 (autocreate support).

---

TLDR; Only adapters which auto create indexes with correct configuration are with this PR:

 - Loupe
 - Memory
 
All others are created with the non correct default settings or will fail and createIndex should definitely still be called for them to update the settings of the specific engines correctly.

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeHelper.php
+++ b/packages/seal-loupe-adapter/src/LoupeHelper.php
@@ -78,21 +78,11 @@ final class LoupeHelper
         }
     }
 
-    public function createIndex(Index $index): Configuration
+    public function createIndex(Index $index): void
     {
-        $indexDirectory = $this->getIndexDirectory($index);
-        if (!\file_exists($indexDirectory)) {
-            \mkdir($indexDirectory, recursive: true);
-        }
-
-        // dumping the configuration allows us to search and index without knowing the configuration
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@Toflar FYI

Toflar:
Thanks!

PR #473, opened 2025-01-18
Add support for Loupe 0.9

alexander-schranz:
This will allow to use also the latest Loupe PHP 0.9 version. There are no changes which effects our code: https://github.com/loupe-php/loupe/releases/tag/0.9.0. Thx to @Toflar

### DIFF
--- a/packages/seal-loupe-adapter/composer.json
+++ b/packages/seal-loupe-adapter/composer.json
@@ -26,13 +26,10 @@
             "email": "alexander@sulu.io"
         }
     ],
-    "conflict": {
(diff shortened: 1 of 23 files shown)


packages/seal-loupe-adapter/composer.json
```diff
@@ -26,13 +26,10 @@
             "email": "alexander@sulu.io"
         }
     ],
-    "conflict": {
-        "doctrine/dbal": "<3.6.0"
-    },
```
alexander-schranz, line 31: conflict not longer required as dbal version already since longer defined correctly now for loupe

PR #474, opened 2025-01-18
Update lock file marking

szepeviktor:
Composer's lock file is ours, it is generated by Composer.

https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#summary

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-composer.lock linguist-vendored
+composer.lock linguist-generated


alexander-schranz:
The problem I see is that `generated` is marked this way:

> Excluded from stats, hidden in diffs

The `hidden in diffs` part is what I don't like so much. Do you know if there is still an option then to see the diff in the PR github view?

szepeviktor:
> Do you know if there is still an option then to see the diff in the PR github view?

Hidden means:
image

Only the content is hidden.

alexander-schranz:
@szepeviktor Thanks!

szepeviktor:
You're welcome.

PR #485, opened 2025-02-20
Fix mezzio linting task and php requirement and fix Yii example run

alexander-schranz:
currently the linting fails because of old definitions in the mezzio composer.json example. this only effects the example.


.examples/yii/config/common/routes.php
```diff
@@ -7,7 +7,7 @@
 use Yiisoft\Router\Route;
 
 return [
-    Group::create('/{_language}')
+    Group::create()
```
alexander-schranz, line 10: changed in the app skeleton example: https://github.com/yiisoft/app/commit/dac04b189868b962f3f7328a55d7f9be9af3f0cd

packages/seal-opensearch-adapter/composer.json
```diff
@@ -41,7 +41,8 @@
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-phpunit": "^1.3",
         "phpunit/phpunit": "^10.3",
-        "rector/rector": "^1.0"
+        "rector/rector": "^1.0",
+        "symfony/http-client": "^5.4 || ^6.0 || ^7.0"
```
alexander-schranz, line 45: else it did automatically add 7.3-x when run `composer lint` adding it directly seems tow ork

.examples/yii/composer.json
```diff
@@ -82,7 +82,7 @@
         "yiisoft/router-fastroute": "^3.0",
         "yiisoft/translator": "^3.0",
         "yiisoft/translator-message-php": "^1.1",
-        "yiisoft/view": "^10.0",
+        "yiisoft/view": "^12.0",
```
alexander-schranz, line 85: required now see update in yii app

.examples/mezzio/composer.json
```diff
@@ -32,7 +32,7 @@
             "phpstan/extension-installer": true
         },
         "platform": {
-            "php": "8.1.16"
+            "php": "8.2.27"
```
alexander-schranz, line 35: it requires 8.2 to work correctly now

PR #497, opened 2025-02-27
Add helper to ease creating an empty result

Toflar:
Creating an empty search result is a very valid case and because creating an empty generator is a bit cumbersome, I think this qualifies for a helper factory method 😊

### DIFF
--- a/packages/seal/src/Search/Result.php
+++ b/packages/seal/src/Search/Result.php
@@ -33,7 +33,7 @@ final class Result extends \IteratorIterator
         return $this->total;
     }
 
-    public static function empty(): static
+    public static function createEmpty(): static
     {
         return new self((static function (): \Generator {
             yield from [];
--- a/packages/seal/src/Search/Result.php
+++ b/packages/seal/src/Search/Result.php
@@ -32,4 +32,11 @@ class Result extends \IteratorIterator
     {
         return $this->total;
     }
+
+    public static function empty(): static
+    {
+        return new static((static function (): \Generator {
+            yield from [];
+        })(), 0);
+    }
 }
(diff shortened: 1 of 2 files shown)


Toflar:
We should probably use `new self()` and make that class `final`. Wdyt?

packages/seal/src/Search/Result.php
```diff
@@ -32,4 +32,11 @@ public function total(): int
     {
         return $this->total;
     }
+
+    public static function empty(): static
```
alexander-schranz: as static function can also be called unstatic way I would prefer `createEmpty`, what do you think?

alexander-schranz:
```suggestion
    public static function createEmpty(): static
```

packages/seal/tests/Search/ResultTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->assertSame(1, $result->total());
+        $this->assertSame([['id' => 42]], \iterator_to_array($result));
+    }
+
+    public function testCreateEmptyResult(): void
+    {
+        $result = Result::empty();
+        $this->assertSame(0, $result->total());
+        $this->assertSame([], \iterator_to_array($result));
+    }
```
alexander-schranz:
```suggestion
    public function testCreateEmptyResult(): void
    {
        $result = Result::createEmpty();
        $this->assertSame(0, $result->total());
        $this->assertSame([], \iterator_to_array($result));
    }
```

alexander-schranz (changes requested):

alexander-schranz:
Docker seems kind of down/flacky so fails unrelated.

alexander-schranz (approved):

PR #498, opened 2025-03-01
Add facet or aggregatable option to schema index field definition

### LINKED ISSUE #182: [Core] Support for Faceting / Aggregations
Typical shops will allow if you search something that they show categories to add additional filters. In that case mostly only categories are shown which matches the current search condition of the user. This kind of things are called aggregations or faceting.

There are different kind of aggregations which would be nice to support but requires a lot of research if all kind of currently supported search engines support such kind of aggregations:

 - category filter aggregations (matching categories with search term)
 - price range filter aggregations (lowest and highest price)
 
###  What kind of search engines support this kind of aggregations
  
Supported:

 - Elasticsearch / Opensearch (https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html @wachterjohannes maybe you can show how currently this was done in such projects). Interesting can also be post filters: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/filter-search-results.html#post-filter 

Maybe:
 
 - RediSearch (not tested yet (https://redis.io/docs/stack/search/reference/aggregations/)
 - Algolia (not tested yet (https://www.algolia.com/doc/guides/managing-results/refine-results/faceting/#retrieving-facets))
 - Meilisearch (not tested yet (https://www.meilisearch.com/docs/learn/advanced/faceted_search#facet-distribution))
 - Solr (not tested yet (https://solr.apache.org/guide/8_8/search-sample.html#facet))
 - Typesense ( not tested yet (https://typesense.org/docs/28.0/api/search.html#facet-results))

 
### Should aggregations be part of the search result or own call?

Some search engines support aggregations as combination with the search query, some already requires that the aggregation is a seperate query. Not sure what from the DX point if `SearchBuilder` allows to query aggregations or if a seperate `AggregationBuilder` would be better and provide easier DX.

### Which type of aggregations are required?

 - Term Aggregation? (e.g.: category filter)
 - Min/Max Aggregation? (e.g.: price range filter)
 - Average Aggregation (show average price value)
 - ...? What other aggregations exists?

alexander-schranz:
### Schema Definition

For ecommerce websites it would great if we could support some kind of facets or aggregations to allow better filtering.

The first case for supporting this is adding it to the schema metadata via:

```php
new Field\TextField(name: 'category', filterable: true, facet: true);
new Field\IntegerField(name: 'downloads', filterable: true, facet: true);
new Field\FloatField(name: 'rating', filterable: true, facet: true);
new Field\DateTimeField(name: 'createdAt', filterable: true, facet: true);
new Field\BooleanField(name: 'isSpecial', filterable: true, facet: true);
```

> [!NOTE]  
> Not sure if `aggregatable` is a better name in the schema index field definition.
>
> ```php
> new Field\TextField(name: 'category', filterable: true, aggregatable: true);
> ```

> [!WARNING]  
> Maybe there need to be no different between `filterable` or `aggregatable` / `facet` > currently all engines seems support facets for filterable fields. If not required we can remove the first two commits.

### Search with Facet Result

The facet should be possible to add to the SearchBuilder:

```php
$result = $this->engine->createSearchBuilder('blog')
    ->addFilter(/* ... */)
    ->addFacet(new TermFacet(field: 'category')) // TODO
    ->addFacet(new CountFacet(field: 'category')) // alternative to Term?
    ->addFacet(new MinMaxFacet(field: 'category')) // TODO
    ->getResult();
```

> [!NOTE]  
> Not sure if `addAggregation` is a better with:
>
> ```php
> ->addAggregation(new TermAggregation(field: 'category')) // TODO
> ->addAggregation(new CountAggregation(field: 'category')) // alternative to Term?
> ->addAggregation(new MinMaxAggregation(field: 'category')) // TODO
> ```

We need maybe know what kind of facets:

# Todo Facets / Aggregations

For the first case I think Term aggregation / facet is best way yet to support categorized search results. Min/Max seems not all engine supported so we will keep that outside currently. AVG / SUM is in Algolia for example return when do facetStats to get min, max, avg, sum but not sure about AVG / SUM as I don't know yet a real usecase for searches and ecommerce sites yet for it.

 - [ ] Term
 - [ ] Min/Max / range ?
 - [ ] AVG / SUM ? (don't know usage about this yet)

# ToDo Decide naming

 - [ ] facet vs. aggregatable

# ToDo Search Engines

 - [ ] Memory
 - [ ] Elasticsearch
 - [ ] Opensearch
 - [ ] Algolia
 - [ ] Meilisearch
 - [ ] Typesense
 - [ ] Solr
 - [ ] Typesense
 - [ ] Loupe

fixes https://github.com/PHP-CMSIG/search/issues/182

### DIFF
--- a/packages/seal-algolia-adapter/src/AlgoliaSchemaManager.php
+++ b/packages/seal-algolia-adapter/src/AlgoliaSchemaManager.php
@@ -92,9 +92,14 @@ final class AlgoliaSchemaManager implements SchemaManagerInterface
             }
         }
 
+        $filterableAndFacetFields = \array_unique([
+            ...$index->filterableFields,
+            ...$index->facetFields,
+        ]);
+
         $attributes = [
             'searchableAttributes' => $index->searchableFields,
-            'attributesForFaceting' => $index->filterableFields,
+            'attributesForFaceting' => $filterableAndFacetFields,
             'replicas' => $replicas,
         ];
 
--- a/packages/seal-elasticsearch-adapter/src/ElasticsearchSchemaManager.php
+++ b/packages/seal-elasticsearch-adapter/src/ElasticsearchSchemaManager.php
@@ -86,12 +86,12 @@ final class ElasticsearchSchemaManager implements SchemaManagerInterface
                 $field instanceof Field\IdentifierField => $properties[$name] = [
                     'type' => 'keyword',
                     'index' => $field->searchable,
-                    'doc_values' => $field->filterable || $field->sortable, // @phpstan-ignore-line
+                    'doc_values' => $field->filterable || $field->sortable || $field->facet, // @phpstan-ignore-line
                 ],
                 $field instanceof Field\TextField => $properties[$name] = \array_replace([
                     'type' => 'text',
                     'index' => $field->searchable,
-                ], ($field->filterable || $field->sortable) ? [
+                ], ($field->filterable || $field->sortable || $field->facet) ? [
                     'fields' => [
                         'raw' => ['type' => 'keyword'],
                     ],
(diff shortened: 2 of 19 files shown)


alexander-schranz:
close in favor of: https://github.com/PHP-CMSIG/search/pull/577

PR #501, opened 2025-03-18
Add support for Loupe 0.10

alexander-schranz:
This will allow to use also the latest Loupe PHP 0.10 version. There are no changes which effects our code: https://github.com/loupe-php/loupe/releases/tag/0.10.0. Thx to @Toflar

### DIFF
--- a/.examples/laravel/composer.lock
+++ b/.examples/laravel/composer.lock
@@ -6444,11 +6444,11 @@
             "dist": {
                 "type": "path",
                 "url": "./../../packages/seal-loupe-adapter",
-                "reference": "b007020e3716d43f4e05732970e2a3295b177bef"
+                "reference": "83c3ec276069d4b985022ab62371ff5b2067e01d"
             },
             "require": {
                 "cmsig/seal": "^0.7",
-                "loupe/loupe": "^0.8 || ^0.9",
+                "loupe/loupe": "^0.8 || ^0.9 || ^0.10",
                 "php": "^8.1",
                 "psr/container": "^1.0 || ^2.0"
             },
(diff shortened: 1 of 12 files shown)


Toflar (approved):

Toflar:
It's a bit tedious with these pre-release 0.x versions, I know. I think we're pretty close to v1.0 though 😊

alexander-schranz:
The failing tests seems be related to changes between 0.9.4 and 0.9.5 version: https://github.com/loupe-php/loupe/compare/0.9.4...0.9.5

Think something broke them in: https://github.com/loupe-php/loupe/pull/155

alexander-schranz:
> It's a bit tedious with these pre-release 0.x versions, I know. I think we're pretty close to v1.0 though 😊

Its fine aslong its just changing the `composer.json` its a easy upgrade 😄 

---

The related tests which fails here:

1. https://github.com/PHP-CMSIG/search/blob/eab260700afb2781ca3275531ce5912a97576d59/packages/seal/src/Testing/AbstractSearcherTestCase.php#L430-L433

Which is something like `tags = "UX" AND tags = "UI"` (documents which have both)

> Returns currently no result instead of the matched one.

2. https://github.com/PHP-CMSIG/search/blob/eab260700afb2781ca3275531ce5912a97576d59/packages/seal/src/Testing/AbstractSearcherTestCase.php#L1113-L1119

Which is something like `tags = "Tech" AND (tags = "UX" OR isSpecial = false)`

> Errors: Loupe\Loupe\Exception\FilterFormatException: Col 35: Error: Expected identical multi attributes within same group,"tags" and "isSpecial" given., got 'isSpecial'

alexander-schranz:
https://github.com/loupe-php/loupe/pull/159 fixes the issues. Retrigger the CI here to check if all works.

PR #503, opened 2025-03-21
Introducing statistics to adapters

Toflar:
Currently, the number of documents in an index is the only thing that I could see is available across the board. Potentially there could also be the index size in bytes etc. in the future. So I decided for a `Statistics` object in our abstraction layer that would allow us to easily add more information in the future.

This is just a quick draft, looking for feedback @alexander-schranz :)

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeAdapter.php
+++ b/packages/seal-loupe-adapter/src/LoupeAdapter.php
@@ -17,6 +17,7 @@ use CmsIg\Seal\Adapter\AdapterInterface;
 use CmsIg\Seal\Adapter\IndexerInterface;
 use CmsIg\Seal\Adapter\SchemaManagerInterface;
 use CmsIg\Seal\Adapter\SearcherInterface;
+use CmsIg\Seal\Adapter\StatisticsInterface;
 
 final class LoupeAdapter implements AdapterInterface
 {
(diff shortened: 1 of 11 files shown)


alexander-schranz:
I would add such methods to the `SearcherInterface` which just "read" things because the idea behind the Indexer and Searcher Interface is the responsibility:

 - Writer: IndexerInterface
 - Reader: SearcherInterface

This make implementation of the ReadWrite and Multi Adapter a lot easier. Maybe WriterInterface and ReaderInterface would have been a better choice, maybe will rethink about that. Even we add the an own Interface for BC reasons the Searcher should implement it for the simplicity of the ReadWrite Adapter and so we not require the Adapter with additional getStatistics.

Do you have any other `Statistics` in mind which we may add later also?

Toflar:
> Do you have any other `Statistics` in mind which we may add later also?

Not for now. For me, this is information about the index/search engine itself. So it's not really searching, that's why I added a new "endpoint".

Toflar:
@alexander-schranz would you prefer to have a simple `countDocuments()` on the `EngineInterface`?

alexander-schranz:
Would go with:

 - `EngineInterface::countDocument(string $index): int`
 - `SearchInterface::count(Index $index): int`

Toflar:
Closing this in favor of https://github.com/PHP-CMSIG/search/pull/561.

PR #504, opened 2025-03-21
Introducing a context trimming helper

Toflar:
This will give developers a handy method to strip long contexts between matches 😎

### DIFF
--- a/packages/seal/composer.json
+++ b/packages/seal/composer.json
@@ -35,7 +35,8 @@
         }
     ],
     "require": {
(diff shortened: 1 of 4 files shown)


Toflar:
I am actually considering to close this PR and publish this as its own package because I can think of other use cases outside of SEAL where this might be useful where I wouldn't want to require the entire SEAL core package just for that...

Toflar:
Closing in favor of https://packagist.org/packages/loupe/context-cropper 😊

PR #507, opened 2025-03-31
Fix getFieldByPath for Object and Typed Fields

### LINKED ISSUE #505: Error in Index::getFieldByPath() ?
Hello,
when trying to query ElasticSearch with something like this `group.id=44`, I got a `FieldByPathNotFoundException` stating that the field did not exist.

The exception is thrown in [Schema/Index.php](/PHP-CMSIG/search/blob/0.7/packages/seal/src/Schema/Index.php) : 
```php
public function getFieldByPath(string $path): AbstractField
{
  $pathParts = \explode('.', $path);
  $fields = $this->fields;

  while (true) {
    $field = $fields[\current($pathParts)] ?? null;

    if ($field instanceof TypedField) {
        $fields = $field->types[\current($pathParts)];
        // next() missing here?
    } elseif ($field instanceof ObjectField) {
        $fields = $field->fields;
        // next() missing here?
    } elseif ($field instanceof AbstractField) {
        return $field;
    } else {
        throw new FieldByPathNotFoundException($this->name, $path);
    }
  }
}
```

`$pathParts` is an array, whose items are returned by the `current()` function. But the index is never incremented by `next()`. Given a `$path` of `group.id`, `group` is searched, then `group` again instead of `id`.

ntoniazzi:
fixes #505

### DIFF
--- a/packages/seal/src/Schema/Index.php
+++ b/packages/seal/src/Schema/Index.php
@@ -90,6 +90,7 @@ final class Index
                 $fields = $field->types[\current($pathParts)];
             } elseif ($field instanceof ObjectField) {
                 $fields = $field->fields;
+                \next($pathParts);
             } elseif ($field instanceof AbstractField) {
                 return $field;
             } else {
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Did a few adjustments to support the Type Field also correctly. Thank you for the reproducer and fix!

alexander-schranz:
Released as 0.7.2

Issue #81, opened 2023-01-19
[New package] ODM/ORM based Datamapping to Document classes


alexander-schranz:
The current abstraction requires json like arrays to be indexed and return that indexes. There should be in future another package build on top of `seal` package which will do mapping from `PHP Classes` so `Document Object` to this array and back to it.

So the `seal` package will be something like what in doctrine is `doctrine/dbal` and the `odm` or whatever we will call it will be what `doctrine/orm` is.

A ODM Impelmentaiton could look like this:

```php
<?php

use Schranz\Search\SEAL\Schema\Field;
use Schranz\Search\SEAL\Schema\Index;

#[Index(name: 'news')]
class News {
    #[Field\IdentifierField('id')]
    private string $id;
    
    #[Field\TextField('title')]
    private string $title;
    
    #[Field\TypedField('header', 'type', [
        'image' => [
            'media' => new Field\IntegerField('media'),
        ],
        'video' => [
            'media' => new Field\TextField('media', searchable: false),
        ],
    ])]
    private array $header;
    
    // ...
}
```

# Open Points

 - [ ] `array` <-> `object` mapping: Find the best way to map array to object maybe look at following components:
    - [Symfony Normalizer and Denormalizer](https://symfony.com/doc/current/components/serializer.html)
    - [Symfony PropertyAccessor](https://symfony.com/doc/current/components/property_access.html)
    - [Symfony ObjectMapper](https://github.com/symfony/symfony/pull/51741) (think does not support array to object)
    - [Doctrine Object Instantiator](https://github.com/doctrine/instantiator)
    - [thephpleague/object-mapper](https://github.com/thephpleague/object-mapper)
    - [JanePHP](https://github.com/janephp/janephp)
    - [CuyZ/Valinor](https://github.com/CuyZ/Valinor)
- [ ] Check what we can keep in mind from @Pierstoval feedback here: https://github.com/PHP-CMSIG/search/discussions/512 

# ToDo

- [ ] Own seal-attributes package which contains the Attributes and AttributeSchemaLoader based on [SchemaLoaderInterface](https://github.com/PHP-CMSIG/search/blob/0.7/packages/seal/src/Schema/Loader/LoaderInterface.php)
- [ ] Own seal-odm package with ObjectManager which wraps the EngineInterface and maps array result to the object and back.
- [ ] Doctrine Implementation Objectmapper? from Entity to our own ODM Documents

PR #509, opened 2025-04-04
Add MongoDB Adapter

alexander-schranz:
This PR is stream lining a first prototype to use MongoDB as a Search Adapter, 

### ToDo

 - [x] Create CI Task
   - [ ] Replace pull_request with pull_request_target
 - [x] Adapter
   - [x] Indexer
   - [x] Searcher
   - [x] SchemaManager
 - [x] Client Creation
   - [ ] Check Database selection
 - [x] SchemaManager
   - [x] Exist
   - [x] Create
      - [ ] Validator for Document Schema?
      - [ ] Create Search Index
        - [ ] Strict Schema
   - [x] Drop
 - [x] Searcher
   - [x] getDocument (by Identifier)
   - [ ] Search
   - [ ] Search with Highlighting
   - [ ] Limit and Offset
   - [x] Equal Condition
   - [x] Equal condition with boolean
   - [x] Equal condition with special string
   - [ ] Multi Equal condition
   - [ ] Equal condition with search condition
   - [x] Not equal condition
   - [x] Greater than condition
   - [x] Greater than equal condition
   - [x] Greater than equal condition multi value
   - [x] Less than condition
   - [x] Less than equal condition
   - [ ] Geo distance condition
   - [ ] Geo bounding box condition
   - [ ] Less than equal condition multi value
   - [x] In condition
   - [x] Not in condition
   - [x] Sort by asc
   - [x] Sort by desc
   - [ ] Searching with nested and or conditions

---

Test this out:

```
git clone git@github.com:PHP-CMSIG/search.git
cd search
git remote add alex git@github.com:alexander-schranz/schranz-search.git
git checkout feature/mongodb-adapter

cd packages/seal-mongodb-adapter
composer install

docker compose up
vendor/bin/phpunit
```

### DIFF
--- a/packages/seal-mongodb-adapter/composer.json
+++ b/packages/seal-mongodb-adapter/composer.json
@@ -1,0 +1,79 @@
+{
+    "name": "cmsig/seal-mongodb-adapter",
+    "description": "An adapter to support MongoDB in cmsig/seal search abstraction.",
+    "type": "library",
+    "license": "MIT",
+    "keywords": [
+        "cmsig",
+        "seal",
+        "seal-adapter",
+        "search-client",
+        "mongodb"
+    ],
+    "autoload": {
+        "psr-4": {
+            "CmsIg\\Seal\\Adapter\\MongoDB\\": "src"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "CmsIg\\Seal\\Adapter\\MongoDB\\Tests\\": "tests"
+        }
+    },
+    "authors": [
+        {
+            "name": "Alexander Schranz",
+            "email": "alexander@sulu.io"
+        }
+    ],
+    "require": {
+        "php": "^8.1",
+        "cmsig/seal": "^0.7",
+        "mongodb/mongodb": "^1.20"
+    },
+    "require-dev": {
+        "php-cs-fixer/shim": "^3.51",
+        "phpstan/extension-installer": "^1.2",
+        "phpstan/phpstan": "^1.10",
+        "phpstan/phpstan-phpunit": "^1.3",
+        "phpunit/phpunit": "^10.3",
+        "rector/rector": "^1.0"
+    },
+    "scripts": {
+        "test": [
+            "Composer\\Config::disableProcessTimeout",
+            "vendor/bin/phpunit"
+        ],
+        "phpstan": "@php vendor/bin/phpstan analyze",
+        "lint-rector": "@php vendor/bin/rector process --dry-run",
+        "lint-php-cs": "@php vendor/bin/php-cs-fixer fix --verbose --diff --dry-run",
+        "lint": [
+            "@phpstan",
+            "@lint-php-cs",
+            "@lint-rector",
+            "@lint-composer"
+        ],
+        "lint-composer": "@composer validate --strict",
+        "rector": "@php vendor/bin/rector process",
+        "php-cs-fix": "@php vendor/bin/php-cs-fixer fix",
+        "fix": [
+            "@rector",
+            "@php-cs-fix"
+        ]
+    },
+    "repositories": [
+        {
+            "type": "path",
+            "url": "./../*",
+            "options": {
+                "symlink": true
+            }
+        }
+    ],
+    "minimum-stability": "dev",
+    "config": {
+        "allow-plugins": {
+            "phpstan/extension-installer": true
+        }
+    }
+}
--- a/packages/seal-mongodb-adapter/src/ClientWrapper.php
+++ b/packages/seal-mongodb-adapter/src/ClientWrapper.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the CMS-IG SEAL project.
+ *
+ * (c) Alexander Schranz <alexander@sulu.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CmsIg\Seal\Adapter\MongoDB;
+
+use MongoDB\Client;
+use MongoDB\Database;
+
+final class ClientWrapper
+{
+    public function __construct(
+        private readonly Client $client,
+        private readonly string $databaseName,
+    ) {
+    }
+
+    public function getClient(): Client
+    {
+        return $this->client;
+    }
+
+    public function getDatabase(): Database
+    {
+        return $this->client->getDatabase($this->databaseName);
+    }
+}
--- a/packages/seal-mongodb-adapter/src/MongoDBAdapter.php
+++ b/packages/seal-mongodb-adapter/src/MongoDBAdapter.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the CMS-IG SEAL project.
+ *
+ * (c) Alexander Schranz <alexander@sulu.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CmsIg\Seal\Adapter\MongoDB;
+
+use CmsIg\Seal\Adapter\AdapterInterface;
+use CmsIg\Seal\Adapter\IndexerInterface;
+use CmsIg\Seal\Adapter\SchemaManagerInterface;
+use CmsIg\Seal\Adapter\SearcherInterface;
+
+final class MongoDBAdapter implements AdapterInterface
+{
+    private readonly SchemaManagerInterface $schemaManager;
+
+    private readonly IndexerInterface $indexer;
+
+    private readonly SearcherInterface $searcher;
+
+    public function __construct(
+        ClientWrapper $client,
+        SchemaManagerInterface|null $schemaManager = null,
+        IndexerInterface|null $indexer = null,
+        SearcherInterface|null $searcher = null,
+    ) {
+        $this->schemaManager = $schemaManager ?? new MongoDBSchemaManager($client);
+        $this->indexer = $indexer ?? new MongoDBIndexer($client);
+        $this->searcher = $searcher ?? new MongoDBSearcher($client);
+    }
+
+    public function getSchemaManager(): SchemaManagerInterface
+    {
+        return $this->schemaManager;
+    }
+
+    public function getIndexer(): IndexerInterface
+    {
+        return $this->indexer;
+    }
+
+    public function getSearcher(): SearcherInterface
+    {
+        return $this->searcher;
+    }
+}
--- a/packages/seal-mongodb-adapter/src/MongoDBAdapterFactory.php
+++ b/packages/seal-mongodb-adapter/src/MongoDBAdapterFactory.php
@@ -1,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the CMS-IG SEAL project.
+ *
+ * (c) Alexander Schranz <alexander@sulu.io>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CmsIg\Seal\Adapter\MongoDB;
+
+use CmsIg\Seal\Adapter\AdapterFactoryInterface;
+use CmsIg\Seal\Adapter\AdapterInterface;
+use MongoDB\Client;
+use Psr\Container\ContainerInterface;
+
+/**
+ * @experimental
+ */
+class MongoDBAdapterFactory implements AdapterFactoryInterface
+{
+    public function __construct(
+        private readonly ContainerInterface|null $container = null,
+    ) {
+    }
+
+    public function createAdapter(array $dsn): AdapterInterface
+    {
+        $client = $this->createClient($dsn);
+
+        return new MongoDBAdapter($client);
+    }
+
+    /**
+     * @internal
+     *
+     * @param array{
+     *     scheme: string,
+     *     host: string,
+     *     port?: int,
+     *     user?: string,
+     *     pass?: string,
+     *     path?: string,
+     *     query: array<string, string>,
+     *     fragment?: string,
+     * } $dsn
+     */
+    public function createClient(array $dsn): ClientWrapper
+    {
+        if ('' === $dsn['host']) {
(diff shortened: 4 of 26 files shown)


packages/seal-mongodb-adapter/src/ClientWrapper.php
```diff
@@ -0,0 +1,36 @@
...
+    public function getClient(): Client
+    {
+        return $this->client;
+    }
+
+    public function getDatabase(): Database
+    {
+        return $this->client->getDatabase($this->databaseName);
+    }
+}
```
alexander-schranz, line 36: Not sure about this wrapper class didn't want that the services always require:

```php
        private readonly Client $client,
        private readonly string $databaseName,
```

and wanted a hande `getDatabase` method to get the current configured database for this "Adapter" instance.

I'm also not sure if my DSN will conflict with MongoDB URI parsing I'm using the DSN `path` part to select the database:

https://github.com/PHP-CMSIG/search/blob/903fe7414b6a168dc7f4758e4a16038dfae13012/packages/seal-mongodb-adapter/src/MongoDBAdapterFactory.php#L75-L85

packages/seal-mongodb-adapter/src/MongoDBSearcher.php
```diff
@@ -0,0 +1,224 @@
...
+        foreach ($filters as $filter) {
+            match (true) {
+                $filter instanceof Condition\IdentifierCondition => $filterQueries[]['_id'][] = $filter->identifier,
+                $filter instanceof Condition\SearchCondition => $filterQueries[]['$search'] =[
+                    'index' => $index->name . '-search-index',
+                    'text' => [
+                        'query' => $filter->query,
+                        'path' => $index->searchableFields,
+                    ],
+                ],
```
alexander-schranz, line 184: Hello @gromnan @alcaeus :wave:,

I greated a quick prototype but where I'm stuck is with the query. I implemented the basic filtering =, !=, <, <=, =>, <, IN and NOT IN filtering. But stuck how combine this now with the search. Maybe you have some hint for me :)

Is there a way to combine the `search` and filters functionality in collection->find ?

packages/seal-mongodb-adapter/src/ClientWrapper.php
GromNaN: The default database can be specified in the "path" part of the DSN: https://www.mongodb.com/docs/manual/reference/connection-string/#connection-string-database-options

The `MongoDB\Client` class doesn't have the notion of default database, you have to extract the default database name from the DSN (or other options).

packages/seal-mongodb-adapter/src/MongoDBAdapter.php
```diff
@@ -0,0 +1,54 @@
...
+final class MongoDBAdapter implements AdapterInterface
+{
+    private readonly SchemaManagerInterface $schemaManager;
+
+    private readonly IndexerInterface $indexer;
+
+    private readonly SearcherInterface $searcher;
+
+    public function __construct(
+        ClientWrapper $client,
```
GromNaN, line 30: You can inject the `MongoDB\Database` object instead; instantiation does not open the connection.

packages/seal-mongodb-adapter/src/MongoDBSchemaManager.php
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        return new SyncTask(null); // TODO wait for index drop
+    }
+
+    public function createIndex(Index $index, array $options = []): TaskInterface|null
+    {
+        // $searchDefinition = $this->createPropertiesMapping($index->fields);
+
+        $this->client->getDatabase()->createCollection($index->name);  // TODO check with mongodb team how we could add stricter schema / validator
```
GromNaN, line 53: You can set the [`validator` option](https://www.mongodb.com/docs/php-library/upcoming/reference/method/MongoDBDatabase-createCollection/#parameters) with a `$jsonSchema`. [Example](https://github.com/GromNaN/symfony-demo/blob/780c9513a1015bf3e4b3d1de605726ab16636004/tests/Functional/Plane/BaseApiTestCase.php#L26-L53):

```php
$database->createCollection('planes', [
    'validator' => [
        '$jsonSchema' => [
            'bsonType' => 'object',
            'required' => ['_id', 'name', 'created_at'],
            'properties' => [
                '_id' => ['bsonType' => 'objectId'],
                'name' => ['bsonType' => 'string'],
                'created_at' => ['bsonType' => 'date'],
                'seats' => [
                    'bsonType' => 'array',
                    'items' => [
                        'bsonType' => 'object',
                        'required' => ['_id', 'number', 'row', 'column'],
                        'additionalProperties' => false,
                        'properties' => [
                            '_id' => ['bsonType' => 'objectId'],
                            'number' => ['bsonType' => 'int'],
                            'row' => ['bsonType' => 'int'],
                            'column' => ['bsonType' => 'int'],
                        ],
                    ],
                ],
            ],
            'additionalProperties' => false,
        ],
    ],
]);
```

packages/seal-mongodb-adapter/src/MongoDBSearcher.php
GromNaN: Search using the Lucene index can only be done with the [`$search` stage](https://www.mongodb.com/docs/atlas/atlas-search/aggregation-stages/search/); it must be the first stage of the aggregation pipeline.
The `text` query operator doesn't use the Lucene index but a the specific "text" index that exists in MongoDB server for a long time before the introduction of Atlas Search.

In order to combine Lucene-base search criteria with basic filters, you have to use the [`compound` search operator](https://www.mongodb.com/docs/atlas/atlas-search/compound/). 

> Is there a way to combine the search and filters functionality in `collection->find` ?

You must use `$collection->aggregate($pipeline)` with a `$search` stage first. Atlas Search is incompatible with `$collection->find($query)`.

packages/seal-mongodb-adapter/src/MongoDBSchemaManager.php
```diff
@@ -0,0 +1,158 @@
...
+    public function dropIndex(Index $index, array $options = []): TaskInterface|null
+    {
+        $this->client->getDatabase()->dropCollection($index->name);
+        $this->client->getDatabase()->getCollection($index->name)->dropSearchIndex($index->name . '-search-index');
+
+        if (!($options['return_slow_promise_result'] ?? false)) {
+            return null;
+        }
+
+        return new SyncTask(null); // TODO wait for index drop
```
GromNaN, line 46: Here is the code I use to query the server until the index is ready: https://github.com/mongodb/mongo-php-library/blob/ae6a3a53d507fe38e447dfde4e77c8efe5724925/tests/SpecTests/SearchIndexSpecTest.php#L54

GromNaN (commented):
It's off to a great start! I've tried to answer your questions.

packages/seal-mongodb-adapter/src/MongoDBSearcher.php
alexander-schranz: I see — so the search has its completely own API over aggregate method. Do you think it make sense for the SEAL to support then `find` and basic search or always go to the `aggregate` method with Atlas Search, would `aggregate` also work if I only have filters without a search query?

alexander-schranz:
@GromNaN Really thank you that helps. Will keep you uptodate and let you know. Only added one question here: https://github.com/PHP-CMSIG/search/pull/509#discussion_r2031020783 The other parts I think is now more doing and trying out.

packages/seal-mongodb-adapter/src/MongoDBSearcher.php
GromNaN: When there is only basic filters, it might be more performant to use `find($query)` or and aggregation with `aggregate([['$match' => $query]])`, both will be the same. But I would start with an implementation that uses only the `$search` stage and operators as this is the target of this package.

packages/seal-mongodb-adapter/src/MongoDBAdapter.php
alexander-schranz: I like the idea, if I can do all over the database I think that would be the best option.

packages/seal-mongodb-adapter/composer.json
```diff
@@ -0,0 +1,79 @@
...
+    "authors": [
+        {
+            "name": "Alexander Schranz",
+            "email": "alexander@sulu.io"
+        }
+    ],
+    "require": {
+        "php": "^8.1",
+        "cmsig/seal": "^0.7",
+        "mongodb/mongodb": "^1.20"
```
alcaeus, line 32: Please use the latest versions, inlcuding 2.0:
```suggestion
        "mongodb/mongodb": "^1.21 || ^2.0"
```

packages/seal-mongodb-adapter/src/ClientWrapper.php
alcaeus: I would try to avoid extracting the database name from the connection string - it would require you to parse the connection string and thus be conformant with the above specification. The default database was used by some tools, but the MongoDB drivers don't leverage it for anything other than authentication.

packages/seal-mongodb-adapter/src/MongoDBAdapterFactory.php
```diff
@@ -0,0 +1,92 @@
...
+     *     host: string,
+     *     port?: int,
+     *     user?: string,
+     *     pass?: string,
+     *     path?: string,
+     *     query: array<string, string>,
+     *     fragment?: string,
+     * } $dsn
+     */
+    public function createClient(array $dsn): ClientWrapper
```
alcaeus, line 52: I would avoid creating the connection string from parts. Since MongoDB drivers support connecting to a range of different topologies, the DSN can't easily split into the parts you used above.

For example, when using a replica set or sharded cluster, you don't have a single `host:port` pair, but rather multiples, with each server using different ports:
`mongodb://host-a:12345,host-b:23456`

Authentication is also not always done using user and password, but also through X509 and other authentication mechanisms. And to top it off, the [Initial DNS Seedlist Discovery specification](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md) allows you to specify hosts and connection options through DNS entries which the driver will then look up - this is where the `mysql+srv://` scheme comes in. That also complicates extracting the database name (see earlier comment) as any connection options can be specified in a TXT record.

Long story short: require users pass you a connection string and let them deal with the rest.

packages/seal-mongodb-adapter/src/MongoDBIndexer.php
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        return new SyncTask(null);
+    }
+
+    public function bulk(Index $index, iterable $saveDocuments, iterable $deleteDocumentIdentifiers, int $bulkSize = 100, array $options = []): TaskInterface|null
+    {
+        $identifierField = $index->getIdentifierField();
+
+        foreach (BulkHelper::splitBulk($saveDocuments, $bulkSize) as $bulkSaveDocuments) {
```
alcaeus, line 77: While the API is a little cumbersome due to it requiring array structures, the [Bulk Write API](https://www.mongodb.com/docs/php-library/current/write/bulk-write/) allows you to combine multiple different write operations (such as insertOne, replaceOne, deleteMany) into a single write. The driver will then take care of batching these writes to not exceed the maximum OP_MSG message size. By creating an ordered bulk write (which is the default), writes will stop after the first error; an unordered write will continue writing despite errors but also without guaranteeing that operations are executed in the specified order.

packages/seal-mongodb-adapter/src/MongoDBSearcher.php
alcaeus: I will add that if you require the latest version of the MongoDB driver, you will have access to the aggregation pipeline builder which gives you a programmatic API to assemble a `$search` stage (and any other stages). For examples, check out the tests for [SearchStage](https://github.com/mongodb/mongo-php-library/blob/v2.x/tests/Builder/Stage/SearchStageTest.php) and [CompoundOperator](https://github.com/mongodb/mongo-php-library/blob/v2.x/tests/Builder/Search/CompoundOperatorTest.php). In essence, these would allow you to generate the correct operator based on the various conditions, then combine them all in a single `must` array for the `compound` operator and pass that to `Stage::search`.

packages/seal-mongodb-adapter/src/MongoDBIndexer.php
alexander-schranz: > While the API is a little cumbersome due to it requiring array structures

You mean the documents itself? It later targetted to be mapped to objects in #81, but currently you can use any normalizer denormalizer to get the objects into that struct. The bulk parameters itself support any kind of iterable which mostly I recommend using \Generators or something which keep memory usage low for reindexing all documents. 

> The driver will then take care of batching these writes to not exceed the maximum OP_MSG message size.

That sounds kind of interesting maybe better in my case would be aslong as possible give the iterables to the underlaying layer and change aslong as not own bulk size is given the underlaying client of that search engine takes care of that.

> an unordered write will continue writing despite errors but also without guaranteeing that operations are executed in the specified order.

Thx for the insights here 🙏 .

alcaeus: > You mean the documents itself?

I was talking about the bulk write API in the MongoDB driver - it requires an array of operations like this:
```php
[
    [ 'deleteMany' => [ $filter ] ],
    [ 'deleteOne'  => [ $filter ] ],
    [ 'insertOne'  => [ $document ] ],
    [ 'replaceOne' => [ $filter, $replacement, $options ] ],
    [ 'updateMany' => [ $filter, $update, $options ] ],
    [ 'updateOne'  => [ $filter, $update, $options ] ],
]
```

As for batching and such, you would create a single bulk write containing all operations (insertions and deletions) and then send that to the server. When assembling the actual bulk write command, the driver will automatically split the write into batches if it exceeds the maximum size supported by MongoDB. While this size is dependent on information returned in the MongoDB handshake, it currently supports 100k operations in a single batch, with the message size limited to 48 MB. Considering this, you can assume that most cases will only need a single round trip and no batching.
If you were to go for that option, creating the operations is basically an `array_map` returning the corresponding `insertOne` or `deleteMany` operation, then combining both lists and passing that to the `Collection::bulkWrite()` method.

PR #513, opened 2025-04-09
Use dedicated bundle extension

silasjoisten:


alexander-schranz:
Currently no plan to split this up. I'm fan of that one class is responsible to bring a libraries functionality to the framework like its done in other frameworks. And thats why I really fan of the `AbstractBundle`. Already did such things before it via some hacks: https://github.com/symfony/symfony/issues/45607 but since 6.1 really like that we can do that.

Still thank you for your effort.

PR #514, opened 2025-04-09
Use phpstan baseline instead of code comments

silasjoisten:


alexander-schranz:
currently not planned to add a baseline as it will just grow. I really like the transparents of the annotations comments inside the code, so I directly see here was something skipped form static code analyzers. Still thank you for your effort.

PR #515, opened 2025-04-09
Add community section to research docs

alexander-schranz:
SEAL should not only be a library for PHP devs it should also be a source for everybody around the topic search. So adding a link to https://opensourceconnections.com/slack which is community created @flaxsearch where search specific topics are discussed, member of Vespa, ParadeDB and others are part of it around ~5600 members are inside this Slack.


alexander-schranz:
@gromNaN @alcaeus maybe also a place where MongoDB can share its latest Atlas Search features

PR #525, opened 2025-05-05
Fix and normalize handling of none matching highlight fields (BC Break)

### LINKED ISSUE #524: Highlighting breaks depending on the query
Using OpenSearch adapter what I can see is if that depending on they query it returns errors, my search is simple:

```
$query = $engine->createSearchBuilder('product')
            //->addSortBy('name', 'asc')
            ->addFilter(new Condition\EqualCondition('token', '5d35b407574d5'))
            ->limit($limit)
            ->offset(($page - 1) * $limit)
        ;
        if ($searchTerm) {
            $query->addFilter(new Condition\SearchCondition('*'.$searchTerm.'*'));
            $query->highlight(['name'], '<mark>', '</mark>');
        }
```
`$searchTerms = ''`
Works but if I remove the if, it doesn't works

`$searchTerms = 'b'`

Uncaught PHP Exception AssertionError: "Expected highlight field to be set." at OpensearchSearcher.php line 143 

`$searchTerms = 'br'`
Here works, results:

Image

`$searchTerms = 'bravecto '`(space at the end)
Breaks

`$searchTerms = 'bravecto de'`
Breaks

`$searchTerms = 'bravecto de '` (space at the end)
Image

`$searchTerms = 'bravecto 4 kg'`
Image

`$searchTerms = 'bravecto 4'` and `$searchTerms = 'bravecto 4 '` (space)
both breaks

The error is always the same:  |CRITICA| REQUES Uncaught PHP Exception AssertionError: "Expected highlight field to be set." at OpensearchSearcher.php line 143

alexander-schranz:
There is inconsistent between how the result is if there are multiple "highlight" fields and the a highlight field does not have a match.

 - Elasticsearch, Opensearch, Solr: Do return nothing if the field has no match and so no highlighting is applied
 - Meilisearch, Algolia, Loupe: Returns the fields value
 
To be consistent I decided to go with the Elasticsearch behaviour because it seems my point of view easier for enduser to write: `$document['_formatted']['description'] ?? $document['description']`.  
As the other way `\str_contains($document['_formatted']['description'], $search->highlightPreTag) ? $document['_formatted']['description'] : null`.

@Toflar what do you think?

Another issue appeared which is #524 that none existing fields did throw an error on elastic and opensearch, we can not be as structure strict as I thought we can here.

## BC Break

Some engines in the past did automatically fallback in the highlighted fields to the original value even if there was no match and nothing to highlight. To normalize this behaviour this mechanic is now in control of you. If you did use Loupe, Meilisearch and Algolia and want the previous behaviour back add fallback yourself:

```diff
$title = $document['_formatted']['title']
+            ?? $document['title']
             ?? '';
```

### DIFF
--- a/docs/search-and-filters/index.rst
+++ b/docs/search-and-filters/index.rst
@@ -413,8 +413,14 @@ The abstraction can also be used to highlight the search term in the result.
         ->getResult();
 
     foreach ($result as $document) {
-        $titleWithHighlight = $document['_formatted']['title'] ?? '';
+        $titleWithHighlight = $document['_formatted']['title']
+            ?? $document['title']
+            ?? '';
     }
+
+If the highlight is applied to multiple fields, only the fields that had a match are returned inside ``_formatted``.
+Fields without a match are returned as ``null``. You might want to use the null coalescing operator (``??``)
+to fall back to the original field, as shown above.
 
 .. note::
 
--- a/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
+++ b/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
@@ -55,13 +55,13 @@ final class AlgoliaSearcher implements SearcherInterface
                 );
             } catch (NotFoundException) {
                 return new Result(
-                    $this->hitsToDocuments($search->index, [], []),
+                    $this->hitsToDocuments($search->index, [], [], $search->highlightPreTag),
                     0,
                 );
             }
 
             return new Result(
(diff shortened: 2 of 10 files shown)


Toflar:
I think it makes sense but should probaby also be mentioned in the docs?

alexander-schranz:
@Toflar Good point, did mention the behaviour in the higlighting documentation.

Toflar (approved):

PR #534, opened 2025-05-08
Improve error message about field names in indexes

gisostallenberg:


packages/seal/src/Schema/Index.php
```diff
@@ -136,9 +136,9 @@ private function getAttributes(array $fields, bool $withoutIdentifierField = fal
             );
 
             \assert(
-                1 === \preg_match('/^([a-z]|[A-Z])\w+$/', $field->name),
+                1 === \preg_match('/^([a-z]|[A-Z])\w+$/', $field->name), // see https://regex101.com/r/xR9G6D/1
                 \sprintf(
-                    'A field named "%s" uses unsupported character in index "%s", supported characters are "a-z", "A-Z", "0-9" and "_".',
+                    'The field name format for field "%s" in index "%s" is unsupported , supported characters are "a-z", "A-Z", "0-9" and "_" and the name must start with a letter.',
```
alexander-schranz: Its not `The field` its still `A field named` as there can be different fields with same inside a nested schema.

```suggestion
                    'A field named "%s" in index "%s" uses unsupported format, supported characters are "a-z", "A-Z", "0-9" and "_" and the name must start with a letter.',
```

alexander-schranz (changes requested):
Thx for the PR 👍  small change request:

PS: Tests need to be adopted.

gisostallenberg:
@alexander-schranz Updated tests

alexander-schranz:
@gisostallenberg thank you!

PR #538, opened 2025-05-09
Improve error message about missing highlighted field in hit

gisostallenberg:


alexander-schranz:
@gisostallenberg was there a case when this error message did appear? This mostly more a hint that some of the search engines might have changed unexpectly and we need to fix such case in SEAL core.

gisostallenberg:
I programmatically marked the fields to highlight (and made a mistake there). When that happens it is unclear about what field it is.

If you'd want to make a clear distinction between pre-search and hit moments, the `SearchBuilder` should check the schema when marking fields to highlight and an exception should be given there.

alexander-schranz (approved):

alexander-schranz:
@gisostallenberg thank you!

Issue #540, opened 2025-05-13
Support for adapter specific Index options


alexander-schranz:
A few search engines like Loupe can have specific options which make sense to be on index level. E.g. typetolarance and so on, but do not make sense as part of abstraction. Such options could be supported maybe via:

```php
new Index(name: 'blog', fields: [/* ... */], options: ['typeTolarance' => 2]);
```

This options are adapter specific so not sure if we should make it possible to define it per adapter:

```php
new Index(name: 'blog', fields: [/* ... */], options: [
    'loupe' => [
        'typeTolarance' => 2
    ],
    'meilisearch' => [
        'localizedAttributes' => [/* ... */], // see https://github.com/PHP-CMSIG/search/issues/510
    ],
]);
```

Not sure yet. Still the adapter schemamanager would be responsible to read from `<adapter_key>` as we give the whole Index instance to the schema manager.

Another question is if options are strict or not. As still there maybe should be possibility for kind of metadata for project specific options like label: https://github.com/PHP-CMSIG/search/issues/539

Doctrine is going the way of have it not adapter specific e.g. `jsonb=true` is configured, databases which do not support it just ignore it and database who are supporting it read it and configure it correctly.

What we should definitly not do is try to abstract this options in any way, @alcaeus warned me about that not try to abstract everything.

Related also to https://github.com/PHP-CMSIG/search/issues/510 /cc @tacman

Issue #539, opened 2025-05-09
Allow to set a label for document fields


gisostallenberg:
A field name follows a specific format, as can be seen [here](https://github.com/PHP-CMSIG/search/pull/534/files#diff-05679ee7d66dc437d725d4bf817b2a6d890eb0d42e8481b2784601b4a8a31b2bR139)

It would be nice to allow to set a label on a field (in a fields options?) and apply that labels to the result document.

e.g.
```php
[
    'Company name' => 'ACME',
]
```
instead of 
```php
[
    'company_name' => 'ACME',
]
```

for 
```php
new Index('companies', [
    'uuid' => new Field\IdentifierField('uuid'),
    'company_name' => new Field\TextField('company_name', options: ['label' => 'Company name']),
]);

**Comment by alexander-schranz** [comment]
Thx for the issue, but currently no plan on implementing such things. Because it just opens more complexity and lot of questions (what is expected format for index method, and if no effect why is search result format different to a index format, if label also used there its a diff between on filters). So lot of things I currently don't want to tackle inside the SEAL core.

And so think such mapper code better should be created inside your projects code instead:

Somebody still can create an own service or helper package based on schema field options to convert 'company_name` to `Company name` based on the metadata. E.g.:

```php
foreach ($document as $oldFieldKey => $fieldValue) {
   $newFieldKey = $index->getFieldByPath($oldFieldKey)->getOptions()['label'] ?? $oldFieldKey;
   \unset($fieldName);
   $document[$oldFieldKey] = $fieldValue;
}
```

Sure need some more handling for nested objects and typed fields but think you get the idea. You even can decorate the EngineInterface if your project always require and give that data.

PR #541, opened 2025-05-13
Add possibility that an Index can have options

alexander-schranz:
Maybe there are adapter specific options for an index. Like we have already `options` array for `fields` we add here it for Index. So the `SchemaManager` of adapters like `Loupe` can example configure a `TypoTolerance` or other options.

fixes part of #540

Target of this PR is only add possibility to have options not yet any
adapter specific implementation for it. It can be used for projects to add additonal infos like a label  like in #539.

### DIFF
--- a/packages/seal/src/Schema/Index.php
+++ b/packages/seal/src/Schema/Index.php
@@ -44,10 +44,12 @@ final class Index
 
     /**
      * @param array<string, AbstractField> $fields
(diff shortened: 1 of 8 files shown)


alexander-schranz:
Why we are not yet sure about if we will solve #540 via index options. Index options will still be added to add possibilities for additonal none core related data like we already do for fields.

PR #545, opened 2025-05-18
Add support for Loupe 0.11 with new offset and limit possibilities

ker0x:


alexander-schranz:
Thx for the pull requests. We will drop support for older Loupe version in favor of supporting the new offset limit functionality and refactor so our searcher to use them. See https://github.com/loupe-php/loupe/pull/188#pullrequestreview-2832420175 do you want to tackle that also?

ker0x:
@alexander-schranz Yes, I can try! I've updated the PR accordingly.

alexander-schranz:
Looks good, thank you @ker0x 

PS: We will probably create a new minor release for this and some upcoming changes. If you require loupe 0.11 before the release you can try `composer require loupe/loupe:"0.11.0 as 0.10.999"`

PR #551, opened 2025-05-26
Add support for distinct attributes

Toflar:
Implements https://github.com/PHP-CMSIG/search/issues/550


packages/seal/src/Schema/Index.php
```diff
@@ -58,6 +60,23 @@ public function __construct(
...
+    {
+        if (null !== $distinctAttribute && !\in_array($distinctAttribute, $this->filterableFields, true)) {
+            throw new \LogicException('The distinct attribute has to be filterable.');
+        }
+
+        $clone = clone $this;
+        $clone->distinctAttribute = $distinctAttribute;
+
+        return $clone;
+    }
```
alexander-schranz: would not add wither methods here we should go the same way as for searchable, filterable, sortable, ... 

I added a deeper description here how I would add such case: https://github.com/PHP-CMSIG/search/issues/550#issuecomment-2910279969

PS: its in SEAL responsible if a field is defined as distinctable and the engine requires it to be also filterable that the schema manager takes core of that.

Toflar: But you cannot have more than one distinct field.

Toflar:
Okay, all seems to work 😎 Loupe fails but I have implemented the logic in https://github.com/loupe-php/loupe/pull/196 already. So I only need to release 0.12 and update to this (maybe in a separate PR)?

The Redis implementation was the most complex because you cannot combine `FT.SEARCH` and `FT.AGGREGATE` apparently. So I worked around that by doing an `FT.AGGREGATE` for identifiers first to apply grouping (which only returns the document IDs) and then `FT.SEARCH` on those to fetch the entire documents. This works perfectly until you have a big pagination setting (I don't know what the max is but if your pagination settings are set to 500, then it will query 500 IDs). Not sure how to tackle that or if we just accept this :)

packages/seal/src/Schema/Index.php
Toflar: I mean, technically you could but in Algolia it has to be known when creating the index.

Toflar:
@alexander-schranz this seems to work for all adapters now 🥳 The linter for Redis drives me nuts, though. But let's have a first review for the concept in general 😊

packages/seal/src/Schema/SchemaHelper.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace CmsIg\Seal\Schema;
+
+class SchemaHelper
```
alexander-schranz: I would keep this internal:

```suggestion
/**
 * @internal
 */
class SchemaHelper
```

But also see later comment I think I would not even use this.

packages/seal-meilisearch-adapter/src/MeilisearchSchemaManager.php
```diff
@@ -58,6 +59,8 @@ public function dropIndex(Index $index, array $options = []): TaskInterface|null
 
     public function createIndex(Index $index, array $options = []): TaskInterface|null
     {
+        SchemaHelper::validateDistinctFieldsAreFilterable($index);
```
alexander-schranz: I understand why you did this, but I would not go this way. We are also in elasticsearch do a lot of `$field->filterable || $field->sortable` to get the settings into the need to support that features, so its fine that we set the field to filterable if the engine requires it to be filterable if distinct is enabled.

```php
'filterableAttributes' => \array_unique(\array_merge(
    $index->filterableFields,
    $index->distinctableFields, // to use distinct the field also need to be filterable
)),
```

packages/seal-algolia-adapter/src/AlgoliaSchemaManager.php
```diff
@@ -80,6 +80,10 @@ public function dropIndex(Index $index, array $options = []): TaskInterface|null
 
     public function createIndex(Index $index, array $options = []): TaskInterface|null
     {
+        if (\count($index->distinctFields) > 1) {
+            throw new \LogicException('Algolia does not support more than one distinct field.');
+        }
```
alexander-schranz: Created an issue to reference this in code so people can may give feedback there:

```suggestion
        if (\count($index->distinctFields) > 1) {
            throw new \LogicException('Algolia does not support more than one distinct field. See https://github.com/PHP-CMSIG/search/issues/557');
        }
```

packages/seal/src/Schema/Field/AbstractField.php
```diff
@@ -27,6 +27,7 @@ public function __construct(
         public readonly bool $searchable,
         public readonly bool $filterable,
         public readonly bool $sortable,
+        public readonly bool $distinct,
```
alexander-schranz, line 30: would have prefereed a `-able` word but hard to find one distinctable and deduplicatable seems not be official words 🤔

alexander-schranz (changes requested):

alexander-schranz:
> The linter for Redis drives me nuts, though. But let's have a first review for the concept in general 😊

Can tackle that then for you.

packages/seal/src/Schema/Field/AbstractField.php
Toflar: I think `distinct` is correct. This is the adjective how you would ask in English:

- "Is this field `searchable`?"
- "Is this field `filterable`?"
- "Is this field `distinct`?"

packages/seal-meilisearch-adapter/src/MeilisearchSchemaManager.php
Toflar: Sure, updated!

packages/seal/src/Schema/SchemaHelper.php
Toflar: Deleted :)

packages/seal/src/Schema/Field/AbstractField.php
alexander-schranz: 👍

packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
```diff
@@ -113,6 +113,10 @@ public function search(Search $search): Result
             ];
         }
 
+        if (null !== $search->distinct) {
+            $body['collapse']['field'] = $search->distinct;
+        }
```
alexander-schranz: depending on: https://www.elastic.co/docs/reference/elasticsearch/rest-apis/collapse-search-results

> The field used for collapsing must be a single valued [keyword](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/keyword) or [numeric](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/number) field with [doc_values](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/doc-values) activated.

We need adopt https://github.com/PHP-CMSIG/search/blob/10ae94639c9027c5f07cb267be89f320c60cdbd6/packages/seal-elasticsearch-adapter/src/ElasticsearchSchemaManager.php#L91-L98:

```diff
                $field instanceof Field\TextField => $properties[$name] = \array_replace([
                    'type' => 'text',
                    'index' => $field->searchable,
-                ], ($field->filterable || $field->sortable) ? [
+                ], ($field->filterable || $field->sortable || $field->distinct) ? [
                    'fields' => [
                        'raw' => ['type' => 'keyword'],
                    ],
                ] : []),
```

Same for Opensearch.

packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
```diff
@@ -113,6 +113,10 @@ public function search(Search $search): Result
             ];
         }
 
+        if (null !== $search->distinct) {
+            $body['collapse']['field'] = $search->distinct;
```
alexander-schranz: If its a text field we need the `.raw` field in opensearch and elasticsearch, the getFilterField helper in the specific searcher handles that:

```suggestion
            $body['collapse']['field'] = $this->getFilterField($this->search->index, $this->search->distinct);
```

Same I think for Opensearch.

Unsure about Solr

alexander-schranz (changes requested):
I think there may be some issue with the text field type and distinct for Elastic and Opensearch I think we need access the .raw field. On Solr I'm not sure how text field are handled there.

alexander-schranz (approved):

alexander-schranz:
Awesome feature, thank you @Toflar!

PR #552, opened 2025-05-27
Allow Loupe v0.12

Toflar:
Not sure what else is needed, the API is identical 😊

### DIFF
--- a/packages/seal-loupe-adapter/composer.json
+++ b/packages/seal-loupe-adapter/composer.json
@@ -29,7 +29,7 @@
     "require": {
         "php": "^8.1",
         "cmsig/seal": "^0.9",
-        "loupe/loupe": "^0.11",
+        "loupe/loupe": "^0.12",
         "psr/container": "^1.0 || ^2.0"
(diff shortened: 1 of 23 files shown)


packages/seal-loupe-adapter/composer.json
```diff
@@ -29,7 +29,7 @@
     "require": {
         "php": "^8.1",
         "cmsig/seal": "^0.9",
-        "loupe/loupe": "^0.11",
+        "loupe/loupe": "^0.12",
```
alexander-schranz, line 32:
```suggestion
        "loupe/loupe": "^0.11 || ^0.12",
```

would add it as additional if there is no breaking change.

PS: `composer update` need to be run to update the lock files.

Toflar: I will need to raise the minimum to 0.12 anyway for https://github.com/PHP-CMSIG/search/pull/551 so I'd prefer to do it here not to pollute the PR with useless diffs :)

PR #561, opened 2025-06-18
Add support for counting documents in index (BC Break  ⚠️)

Toflar:
Because this is a BC break, this shouldn't go to 0.9 but I would need a target branch then 😊 

Example:

```php
if ($this->engine->countDocuments('my_index') > 0) {
     // show search
}
```

## BC Breaks

The `EngineInterface` and `SearcherInterface` changed and require each additional method now if you did use them:

```diff
class YourClass implements EngineInterface {
    // ...
    
+    public function countDocuments(string $index): int
+    {
+        return $this->decoratedEngine->countDocuments($index);
+    }
}
```

```diff
class YourClass implements SearcherInterface {
    // ...
    
+    public function count(Index $index): int
+    {
+        // implement counting
+    }
}
```

Most project not decorating or using the interfaces are not effected by this BC-break and can update without any changes.


Toflar:
@alexander-schranz not sure how to test this properly because the statistics are very often not live (e.g. for Meilisearch) - hence it fails if you do not wait long enough.

alexander-schranz:
@Toflar interesting. Thought the stats should be uptodate after the task is complete 🤔

I'm thinking of to add a helper method to the AbstractTestCases which can be overwritten by the specific adapter test classes.

```php
protected function waitForStats(): void
{
   // some adapters can overwrite this method to wait until some tasks are finished.
}
```

```diff
   public function testCount(): void
    {
        $schema = self::getSchema();
        $this->assertSame(0, self::$searcher->count($schema->indexes[TestingHelper::INDEX_COMPLEX]));

        $documents = TestingHelper::createComplexFixtures();

        foreach ($documents as $document) {
            self::$taskHelper->tasks[] = self::$indexer->save(
                $schema->indexes[TestingHelper::INDEX_COMPLEX],
                $document,
                ['return_slow_promise_result' => true],
            );
        }
        self::$taskHelper->waitForAll();
+       $this->waitForStats();

        $this->assertSame(4, self::$searcher->count($schema->indexes[TestingHelper::INDEX_COMPLEX]));

        foreach ($documents as $document) {
            self::$taskHelper->tasks[] = self::$indexer->delete(
                $schema->indexes[TestingHelper::INDEX_COMPLEX],
                $document['uuid'],
                ['return_slow_promise_result' => true],
            );
        }
        self::$taskHelper->waitForAll();
    }
```

Could think of that we in case of meilisearch check if `isIndexing` is false? https://www.meilisearch.com/docs/reference/api/stats

What you think?

packages/seal/src/Testing/AbstractAdapterTestCase.php
```diff
@@ -93,6 +93,8 @@ public function testDocument(): void
         $task = $engine->createSchema(['return_slow_promise_result' => true]);
         $task->wait();
 
+        $this->assertSame(0, $engine->countDocuments(TestingHelper::INDEX_COMPLEX));
```
alexander-schranz: would remove this from the `testDocument` test and move into an own `countDocuments` test.

alexander-schranz: Makes implement new adapters easier, as I mosly start with simple implementation for testDocument case when creating a new adapter and want not tackle count in the same step.

Toflar: Done 👍

Toflar:
> What you think?

Hmm, I noticed that maybe something with the testsuite is wrong in general?

1. `cd packages/seal-meilisearch-adapter`
2. `docker compose up -d`
3. `vendor/bin/phpunit` <-- runs all the tests and `MeilisearchSearcherTest::testCount` fails ❌ 

However, if I run

`vendor/bin/phpunit --filter MeilisearchSearcherTest::testCount` it passes ✅ 

Maybe some other test does not properly reset which would be a bug in the current branch already? Independently of this PR?

alexander-schranz:
Good catch, I merged #563 you can rebase this one.

Toflar:
I'm giving up on fixing those tests.

packages/seal/src/Testing/AbstractAdapterTestCase.php
```diff
@@ -37,6 +37,9 @@ protected function setUp(): void
 
     protected function tearDown(): void
     {
+        $task = self::getEngine()->dropSchema(['return_slow_promise_result' => true]);
+        $task->wait();
```
alexander-schranz:
```suggestion
        try {
            $task = self::getEngine()->dropSchema(['return_slow_promise_result' => true]);
            $task->wait();
        } catch (\Exception) {
            // ignore eventuell not existing indexes to drop
        }
```

alexander-schranz:
@Toflar you were so close 🤏

packages/seal-redisearch-adapter/src/RediSearchSearcher.php
```diff
@@ -39,6 +39,13 @@ public function __construct(
         );
     }
 
+    public function count(Index $index): int
+    {
+        $info = $this->client->rawCommand('FT.INFO', $index->name);
+
+        return (int) $info[9] ?? 0;
```
alexander-schranz: I will have a look at this. Redis is here really tricky about different versions and where things are returned.

alexander-schranz:
@toflar do you want to add a section about counting to the docs here: https://php-cmsig.github.io/search/search-and-filters/index.html#distinct-fields

PR #576, opened 2025-07-21
Allow symfony/console:^7.0 in mezzio integration

kynx:
It looks like the Symfony integration already allows version 7.


alexander-schranz:
Thx for the pull request.

PS: The mezzio integration does not depend on the symfony integration, the version supported is defined by laminas-cli version https://github.com/laminas/laminas-cli but as they seems to have released a Symfony 7 support we can now also allow Symfony 7 in the mezzio integration.

PR #577, opened 2025-07-22
Add support for facets (BC Break ⚠️) 

### LINKED ISSUE #182: [Core] Support for Faceting / Aggregations

Toflar:
## Example

Facetting
---------

Implements `MinMaxFacet` and `CountFacet` which is what most search engines support:

```php
use CmsIg\Seal\Search\Facet\MinMaxFacet;
use CmsIg\Seal\Search\Facet\CountFacet;

$result = $this->engine->createSearchBuilder('blog')
    ->addFacet(MinMaxFacet('rating'))
    ->addFacet(CountFacet('tags'))
    ->getResult();

$facets = $result->facets(); // Output depends on the facet type
```
    
>  For `->addFacet()` to work, your fields (`age` and `tags` in our our example) have to be configured using `facet: true` in the index schema.

> Facets on fields that are configured `multiple` is currently not supported by `RediSearch` adapter.
    See `this GitHub Issue <https://github.com/PHP-CMSIG/search/issues/583>`__ for more information.

<details>

<summary>

   <h3>The Idea</h3>
   
</summary>

Copied from <a href="https://github.com/PHP-CMSIG/search/pull/498">https://github.com/PHP-CMSIG/search/pull/498</>:

### Schema Definition

For ecommerce websites it would great if we could support some kind of facets or aggregations to allow better filtering.

The first case for supporting this is adding it to the schema metadata via:

```php
new Field\TextField(name: 'category', filterable: true, facet: true);
new Field\IntegerField(name: 'downloads', filterable: true, facet: true);
new Field\FloatField(name: 'rating', filterable: true, facet: true);
new Field\DateTimeField(name: 'createdAt', filterable: true, facet: true);
new Field\BooleanField(name: 'isSpecial', filterable: true, facet: true);
```

> [!NOTE]  
> Not sure if `aggregatable` is a better name in the schema index field definition.
>
> ```php
> new Field\TextField(name: 'category', filterable: true, aggregatable: true);
> ```

> [!WARNING]  
> Maybe there need to be no different between `filterable` or `aggregatable` / `facet` > currently all engines seems support facets for filterable fields. If not required we can remove the first two commits.

### Search with Facet Result

The facet should be possible to add to the SearchBuilder:

```php
$result = $this->engine->createSearchBuilder('blog')
    ->addFilter(/* ... */)
    ->addFacet(new TermFacet(field: 'category')) // TODO
    ->addFacet(new CountFacet(field: 'category')) // alternative to Term?
    ->addFacet(new MinMaxFacet(field: 'category')) // TODO
    ->getResult();
```

> [!NOTE]  
> Not sure if `addAggregation` is a better with:
>
> ```php
> ->addAggregation(new TermAggregation(field: 'category')) // TODO
> ->addAggregation(new CountAggregation(field: 'category')) // alternative to Term?
> ->addAggregation(new MinMaxAggregation(field: 'category')) // TODO
> ```

We need maybe know what kind of facets:

# Todo Facets / Aggregations

For the first case I think Term aggregation / facet is best way yet to support categorized search results. Min/Max seems not all engine supported so we will keep that outside currently. AVG / SUM is in Algolia for example return when do facetStats to get min, max, avg, sum but not sure about AVG / SUM as I don't know yet a real usecase for searches and ecommerce sites yet for it.

 - [x] Term
 - [x] Min/Max / range ?
 - [ ] AVG / SUM ? (don't know usage about this yet)

# ToDo Decide naming

 - [x] facet vs. aggregatable

# ToDo Search Engines

 - [x] Memory
 - [x] Elasticsearch
 - [x] Opensearch
 - [x] Algolia
 - [x] Meilisearch
 - [x] Typesense
 - [x] Solr
 - [x] Typesense
 - [x] Loupe
 - [ ] Redis (partial)
 
</details>

fixes https://github.com/PHP-CMSIG/search/issues/182

# BC Breaks

All Schema Field classes constructor methods have an additional argument `facet` before the `options` argument.


Toflar:
@alexander-schranz two questions regarding this PR:

1. Do you think we should try to somehow standardize the `$result->facets()` result? I mean, it is standardized already of course but I mean with some objects. I'm not sure we should do that already because who knows what's going to change here...
2. Regarding https://github.com/PHP-CMSIG/search/issues/583: What do you think about an `$engine->supportsFacet($facet)`? Then people could build their implementations/UIs for different adapters. Basically, if you use Redis you won't have facet X but if you use another adapter, you do 😊

alexander-schranz:
> Do you think we should try to somehow standardize the $result->facets() result? I mean, it is standardized already of course but I mean with some objects. I'm not sure we should do that already because who knows what's going to change here...

Currently not yet with object. As not sure if I will move that kind of things to `ODM` package. So lets keep here the standardized array result.

> Regarding https://github.com/PHP-CMSIG/search/issues/583: What do you think about an $engine->supportsFacet($facet)? Then people could build their implementations/UIs for different adapters. Basically, if you use Redis you won't have facet X but if you use another adapter, you do 😊

Not yet inside this pull request. Will need to think about it as its not the only kind of features which are not supported by all adapters. Created a issue for it: https://github.com/PHP-CMSIG/search/issues/584

packages/seal/src/Testing/AbstractSearcherTestCase.php
```diff
@@ -117,6 +119,75 @@ public function testDistinctSearch(): void
         }
     }
 
+    public function testFacetSearch(): void
```
alexander-schranz: Can we split this test into multiple single tests, make implementing such features in new adapters easier:

 - `testCountFacet` (CountFacet: rating)
 - `testCountFacetOnMultiValue` (CountFacet: tags)
 - `testMinMaxFacet` (MinMaxFacet: rating)
 - `testSearchConditionWithMultipleFacets` (Search, MinMaxFacet: rating, CountFacet: tags)

Toflar: Will do but this now also means work for Loupe because Loupe does not support count facets on `rating`.

alexander-schranz:
PS: CI error is because I already prepared the 0.11 branch, which require a rebase of the branch.

packages/seal/src/Testing/AbstractSearcherTestCase.php
alexander-schranz: I see we can also skip the test for Loupe like we do skip the others for Redis.

Toflar: Nah, I just added support in Loupe and tagged a new version, all good 😊

alexander-schranz: That was fast 🚀 nice.

Toflar:
I think all the remaining issues are unrelated to this PR, so marking this ready for review 😊

alexander-schranz (approved):

alexander-schranz:
Looks good for me 🚀 . Lets merge it.

PR #578, opened 2025-07-22
Add LoaderProviderInterface so different loaders can be configured

kynx:
I'm implementing SEAL in a multi-tenant environment with identical indexes prefixed with the tenant name. Following on from the suggestion at https://github.com/PHP-CMSIG/search/discussions/266 I plan to write a custom `LoaderInterface` implementation to handle this. However, with the Mezzio integration `PhpFileLoader` is hard-coded in the `@internal` `SealContainerFactory`.

This PR adds a `LoaderProviderInterface` and by default maps it to a `PhpFileLoaderProvider`, which follows the same logic as before. Now weirdos like me can override it if needed 😁

### DIFF
--- a/integrations/mezzio/src/ConfigProvider.php
+++ b/integrations/mezzio/src/ConfigProvider.php
@@ -29,11 +29,27 @@ use CmsIg\Seal\Adapter\Typesense\TypesenseAdapterFactory;
 use CmsIg\Seal\EngineInterface;
 use CmsIg\Seal\EngineRegistry;
 use CmsIg\Seal\Integration\Mezzio\Service\CommandAbstractFactory;
+use CmsIg\Seal\Integration\Mezzio\Service\LoaderProviderInterface;
+use CmsIg\Seal\Integration\Mezzio\Service\PhpFileLoaderProvider;
 use CmsIg\Seal\Integration\Mezzio\Service\SealContainer;
 use CmsIg\Seal\Integration\Mezzio\Service\SealContainerFactory;
 use CmsIg\Seal\Integration\Mezzio\Service\SealContainerServiceAbstractFactory;
 use CmsIg\Seal\Schema\Schema;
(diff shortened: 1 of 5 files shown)


kynx:
Hrm, don't quite understand why it fails - I'm have a hell of a job getting the tests going locally because of a `no APPLICATION_ID or ADMIN_API_KEY provided` trying to create the Algolia client. And hints on what I should set those to?

alexander-schranz:
I'm still not sure about this as decorating the PHPFileLoader should still be possible.

Still to run the tests set the `ALGOLIA_DSN` env variable.
You require an algolia account which provides you the information for the DSN:

```
ALGOLIA_DSN=algolia://<APPLICATION_ID>:<ADMIN_API_KEY>
```

See https://www.algolia.com/doc/libraries/php/v4/ and https://dashboard.algolia.com/account/api-keys

kynx:
Thanks @alexander-schranz . The idea here is exactly to allow `SealContainerFactory` to use a decorated loader. Right now it's doing a `new PhpFileLoader()` so I can't see how that's possible: https://github.com/PHP-CMSIG/search/blob/9ab7dd2ad89732fcbf6c112aa5af6f4c969fbfff/integrations/mezzio/src/Service/SealContainerFactory.php#L96. But maybe I'm missing something.

kynx:
Turns out I don't really need the multiple adapter config, and there's other stuff that requires some decoration, so it's easier for me to write my own factories. Thanks.

alexander-schranz:
I will need to have a deeper look at decorating SEAL services, why it seems to be easy in Symfony projects. I need to get more familiar why other dependency injection containers have here some problems. I plan as soon as I got some time to provide a documentation about the decorate PHPFileLoader, if you have other parts still let me know as it sure want to provide documentation examples for such overrides also.

PR #580, opened 2025-07-24
Add searchable and filterable fields during replicate creation in Algolia

bobbyzholmes:
added searchable and filterable fields to the replicate creation


alexander-schranz:
Thank you @bobbyzholmes I pushed some changes so the `geo` field is correctly replaced for the replica and the `attributeForDistinct` also set in the replica.

PR #594, opened 2025-08-03
Fix issues with filters by Date Fields

### LINKED ISSUE #587: Filters not working like expected for date fields
Currently the Loupe adapter stores dates as timestamps, whereas the Elasticsearch adapter uses its native date field. Unless I'm mistaken, this means the same `Search` cannot work with both.

It would be nice to provide an abstraction for this. I suggest using PHP's `DateTimeImmutable` and converting to the required format in the `SearcherInterface` implementation. Thoughts?

alexander-schranz:
Filters based on `datetime` fields seems currently not work correctly in:

 - [x] Loupe (expect integer in filter)
 - [x] Typesense (expect integer in filter)
 - [x] Meilisearch (expect integer in filter, and schema as integer) (requires drop reindex)
 - [x] Algolia (expect integer in filter, and schema as integer) (requires drop reindex)
 - [x] RediSearch (expect integer in filter, and schema as integer) (requires drop reindex)
 - [x] Solr (expect specific format in filters, and in indexing) (requires drop reindex)

⚠️ `Marshaller` and `Flattenmarshaller` services also changed the `dateAsInteger: true` was removed in favor of `dateFormat: 'U'`. To allow different date formats for Example for Solr which requires a specific format. As they are still `@internal` there are no issues with BC Break here. But still a reindex is required for Solr, Redisearch and 

fixes #587

### BC Breaks

For `Meilisearch`, `Algolia`, `RediSearch` and `Solr` the saving of the date time type changed and requires a drop reindex. `cmsig:seal:reindex --drop`.

### DIFF
--- a/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
+++ b/packages/seal-algolia-adapter/src/AlgoliaSearcher.php
@@ -34,6 +34,7 @@ final class AlgoliaSearcher implements SearcherInterface
         private readonly SearchClient $client,
     ) {
         $this->marshaller = new Marshaller(
+            dateFormat: 'U',
             geoPointFieldConfig: [
                 'name' => '_geoloc',
                 'latitude' => 'lat',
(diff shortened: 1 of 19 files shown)


packages/seal-meilisearch-adapter/src/MeilisearchIndexer.php
```diff
@@ -29,6 +29,7 @@ public function __construct(
         private readonly Client $client,
     ) {
         $this->marshaller = new Marshaller(
+            dateFormat: 'U',
```
alexander-schranz, line 32: if I understood correctly https://www.meilisearch.com/docs/learn/filtering_and_sorting/working_with_dates

sorting works by date strings, but for filters we need save a unixtimestamp. So I use like in Loupe and Typesense a UnixTimestap now.

packages/seal-solr-adapter/src/SolrIndexer.php
```diff
@@ -29,6 +29,7 @@ public function __construct(
         private readonly Client $client,
     ) {
         $this->marshaller = new FlattenMarshaller(
+            dateFormat: 'Y-m-d\TH:i:s\Z',
```
alexander-schranz, line 32: Solr Expect a specific date format, While it did work with our `c` format it did remove the timezone incorrectly and so the filters didn't work for exact matches. We are now converting it into correct values.

packages/seal-algolia-adapter/src/AlgoliaSearcher.php
```diff
@@ -34,6 +34,7 @@ public function __construct(
         private readonly SearchClient $client,
     ) {
         $this->marshaller = new Marshaller(
+            dateFormat: 'U',
```
alexander-schranz, line 37: same as for Meilisearch the Algolia requires a unixtimestamp for filtering: https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-attributes/#filter-by-date

PR #603, opened 2025-09-01
Remove none found identifiers in reindex

### LINKED ISSUE #472: Delete IDs of non existent documents in Engine::reindex()
When you pass identifiers in the `ReindexConfig`, the ones that do not exist anymore should get deleted.
Something along the lines of this:

```diff
Subject: [PATCH] Delete reindex identifiers that do not exist anymore
---
Index: packages/seal/src/Engine.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/seal/src/Engine.php b/packages/seal/src/Engine.php
--- a/packages/seal/src/Engine.php	(revision 4ea343448a96c12010b2902b9c6c124c88555d88)
+++ b/packages/seal/src/Engine.php	(date 1736179896990)
@@ -140,16 +140,23 @@
     ): void {
         /** @var array<string, ReindexProviderInterface[]> $reindexProvidersPerIndex */
         $reindexProvidersPerIndex = [];
+        /** @var array<string, string> $identifiersPerIndex */
+        $identifiersPerIndex = [];
         foreach ($reindexProviders as $reindexProvider) {
             if (!isset($this->schema->indexes[$reindexProvider::getIndex()])) {
                 continue;
             }
 
+            $identifiersPerIndex[$reindexProvider::getIndex()] = $this->schema->indexes[$reindexProvider::getIndex()]->getIdentifierField()->name;
+
             if ($reindexProvider::getIndex() === $reindexConfig->getIndex() || null === $reindexConfig->getIndex()) {
                 $reindexProvidersPerIndex[$reindexProvider::getIndex()][] = $reindexProvider;
             }
         }
 
+        // Track documents that need to be deleted if an identifiers array was given
+        $documentIdsToDelete = array_flip($reindexConfig->getIdentifiers());
+
         foreach ($reindexProvidersPerIndex as $index => $reindexProviders) {
             if ($reindexConfig->shouldDropIndex() && $this->existIndex($index)) {
                 $task = $this->dropIndex($index, ['return_slow_promise_result' => true]);
@@ -164,7 +171,7 @@
             foreach ($reindexProviders as $reindexProvider) {
                 $this->bulk(
                     $index,
-                    (function () use ($index, $reindexProvider, $reindexConfig, $progressCallback) {
+                    (function () use ($index, $reindexProvider, $reindexConfig, $progressCallback, $documentIdsToDelete, $identifiersPerIndex) {
                         $count = 0;
                         $total = $reindexProvider->total();
 
@@ -172,6 +179,9 @@
                         foreach ($reindexProvider->provide($reindexConfig) as $document) {
                             ++$count;
 
+                            // Document still exists, do not delete
+                            unset($documentIdsToDelete[$document[$identifiersPerIndex[$index]]]);
+
                             yield $document;
 
                             if (null !== $progressCallback
@@ -193,5 +203,9 @@
                 );
             }
         }
+
+        if ([] !== $documentIdsToDelete) {
+            $this->bulk('???', [], $documentIdsToDelete, $reindexConfig->getBulkSize());
+        }
     }
 }
```

As you can see from the `???` in the patch, I was unsure about the index name. There seems to be a conceptual issue in this method because somehow you pass an array of `$reindexProviders` per index and also the `ReindexConfig` has an `index`. How would you know which IDs in the `ReindexConfig` belong to which index?

alexander-schranz:
To make removal easier the reindex will automatically remove document which where not provided by requested by the ReindexConfig. This make specially handling inside ORMs or Listeners on Add, Update, Remove a lot easier. As they just need to feel up identifiers. Thx goes to @Toflar who provided suggestion in #472.

fixes #472

### ToDo

 - [x] Add test case
 - [x] Reindex is async and requires in some cases return a task

### DIFF
--- a/packages/seal/src/Engine.php
+++ b/packages/seal/src/Engine.php
@@ -145,15 +145,22 @@ final class Engine implements EngineInterface
     ): void {
         /** @var array<string, ReindexProviderInterface[]> $reindexProvidersPerIndex */
         $reindexProvidersPerIndex = [];
+        /** @var array<string, string> $identifiersPerIndex */
+        $identifiersPerIndex = [];
         foreach ($reindexProviders as $reindexProvider) {
             if (!isset($this->schema->indexes[$reindexProvider::getIndex()])) {
                 continue;
             }
 
+            $identifiersPerIndex[$reindexProvider::getIndex()] = $this->schema->indexes[$reindexProvider::getIndex()]->getIdentifierField()->name;
+
             if ($reindexProvider::getIndex() === $reindexConfig->getIndex() || null === $reindexConfig->getIndex()) {
                 $reindexProvidersPerIndex[$reindexProvider::getIndex()][] = $reindexProvider;
             }
         }
+
+        // Track documents that need to be deleted if an identifiers array was given
+        $documentIdsToDelete = array_flip($reindexConfig->getIdentifiers());
 
         foreach ($reindexProvidersPerIndex as $index => $reindexProviders) {
             if ($reindexConfig->shouldDropIndex() && $this->existIndex($index)) {
@@ -169,13 +176,16 @@ final class Engine implements EngineInterface
             foreach ($reindexProviders as $reindexProvider) {
                 $this->bulk(
                     $index,
-                    (function () use ($index, $reindexProvider, $reindexConfig, $progressCallback) {
+                    (function () use ($index, $reindexProvider, $reindexConfig, $progressCallback, $documentIdsToDelete, $identifiersPerIndex) {
                         $count = 0;
                         $total = $reindexProvider->total();
 
                         $lastCount = -1;
                         foreach ($reindexProvider->provide($reindexConfig) as $document) {
                             ++$count;
+
+                            // Document still exists, do not delete
+                            unset($documentIdsToDelete[$document[$identifiersPerIndex[$index]]]);
 
                             yield $document;
 
@@ -197,6 +207,12 @@ final class Engine implements EngineInterface
                     $reindexConfig->getBulkSize(),
                 );
             }
+        }
+
+        if ([] !== $documentIdsToDelete) {
+            $index = $reindexConfig->getIndex();
+            \assert(null !== $index, 'Index must be set if identifiers are given in reindex config.');
+            $this->bulk($index, [], $documentIdsToDelete, $reindexConfig->getBulkSize());
         }
     }
 }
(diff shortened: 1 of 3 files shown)


packages/seal/src/Engine.php
```diff
@@ -198,5 +208,11 @@ public function reindex(
                 );
             }
         }
+
+        if ([] !== $documentIdsToDelete) {
+            $index = $reindexConfig->getIndex();
+            \assert(null !== $index, 'Index must be set if identifiers are given in reindex config.');
```
alexander-schranz, line 222: I think we need force this. Reindex config with identifiers but without index does not make any sense in my opinion.

@Toflar what do you think?

Toflar: Yes, that's what I was saying in the original issue:

> There seems to be a conceptual issue in this method because somehow you pass an array of $reindexProviders per index and also the ReindexConfig has an index. How would you know which IDs in the ReindexConfig belong to which index?

To be honest, I think we should break the API here and not accept any `index` in `ReindexConfig` and also not an array of `$reindexProviders` that are grouped per  `index`.
All the other methods in `Engine` are per index and you always have to provide `string $index`.
I think it should be the same for `reindex()`. Then we would also not have this conceptual problem here.

If anybody wants to `reindex()` over multiple indexes they can call `reindex(string $index...)` multiple times themselves, no?

alexander-schranz: @Toflar the why `reindexProviders` are given here and not `string index` or even `string indexes` is because other frameworks did not have the a concept of `lazy services` and I wanted to avoid that instancing the Engine instances just for searching creates all reindex providers service with all its dependencies (ORMs, ...).

But yes think reindexProviders would make more sense as part of the engine constructor as part of that method, maybe an own Lazy implementation for such frameworks via a callback can be implemented. Need to have a look at the different frameworks integration. For Symfony `lazy="true"` for all providers is I think the easist.

PS: We still need a way to reindex all via a command.

But we could currently still implement this and move that refactoring into an own Issue or Pull Request.

alexander-schranz: Created a proposal here: https://github.com/PHP-CMSIG/search/issues/604

Toflar: I guess we could also provide a `LazyEngine` that implements the `EngineInterface` maybe 🤔

alexander-schranz:
Merging this the refactoring of reindex is an own topic. (#604)

PR #609, opened 2025-09-25
Fix #607 typo in documentation

### LINKED ISSUE #607: [Documentation] Invalid readme title
This file incorrectly states that it's a readme for another adapter, just so you're aware. (Meilisearch instead of Opensearch)

https://github.com/PHP-CMSIG/search/blob/0.12/packages/seal-opensearch-adapter/README.md

lochmueller:


alexander-schranz:
@lochmueller thank you!

PR #620, opened 2025-11-18
Add virtual packages for seal adapters

alexander-schranz:
All seal adapters will provide `cmsig/seal-adapter-implementation`.

All seal adapters in core go with:

```json
    "provide": {
        "cmsig/seal-adapter-implementation": "self.version"
    },
```

Custom adapters need provide the related / supported `cmsig/seal` version:

```json
    "provide": {
        "cmsig/seal-adapter-implementation": "0.12.0"
    },
```

As I personally not a big fan of the virtual packages we will currently not require them in our `integration` packages. But third party are this way free to require it to make composer fail if no adapter was required by the project / library.

This should help eventuall a future `contao/backend-search-bundle` package to require `"cmsig/seal-adapter-implementation": "^0.12"` if they want that behaviour.

### DIFF
--- a/.examples/laravel/composer.lock
+++ b/.examples/laravel/composer.lock
@@ -8,25 +8,25 @@
     "packages": [
         {
             "name": "brick/math",
-            "version": "0.13.1",
+            "version": "0.14.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/brick/math.git",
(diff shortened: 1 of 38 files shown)


Toflar (approved):
I don't think we're going to use it in Contao after all but in any case, I think this is the right step for whoever builds a package that depends on SEAL and thus on at least one adapter 👍

PR #623, opened 2025-12-03
Update PHPStan, Rector to Version 2 also with Symfony 7.4

alexander-schranz:
Symfony 7.4 was released last week, so update all dependencies to newer versions.


alexander-schranz:
The Lint task currently failing because of: https://github.com/symfony/symfony/issues/62629

PR #629, opened 2026-01-07
Fix Elasticsearch Lexical Error in SearchCondition

botzkobg:
The elasticsearch adapter converts `SearchCondition` to `query_string` filter. This can be used for very powerful searches but is also very fragile. If the search functionality is presented to a user that doesn't know the search syntax and structure or is not aware that elasticsearch database is used this could lead to errors.

This PR switches from  `query_string` to `multi_match`. This will remove the powerful search provided by `query_string` but will also remove all errors that will be produced by the simple user. It also should most closely match the search functionality to all other adapters that doesn't have the same capabilities like the elasticsearch `query_string` filter.

[Specific characters](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query#_reserved_characters) used by `query_string`
[Warning](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query) to use `match` instead of `query_string` if the extra functionality is not required]
`multi_match` [docs](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-multi-match-query)

### DIFF
--- a/packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
+++ b/packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
@@ -243,7 +243,7 @@ final class ElasticsearchSearcher implements SearcherInterface
         foreach ($filters as $filter) {
             match (true) {
                 $filter instanceof Condition\IdentifierCondition => $filterQueries[]['ids']['values'][] = $filter->identifier, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
-                $filter instanceof Condition\SearchCondition => $filterQueries[]['multi_match']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
+                $filter instanceof Condition\SearchCondition => $filterQueries[]['bool']['must']['multi_match']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
                 $filter instanceof Condition\EqualCondition => $filterQueries[]['term'][$this->getFilterField($index, $filter->field)]['value'] = $filter->value, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
                 $filter instanceof Condition\NotEqualCondition => $filterQueries[]['bool']['must_not']['term'][$this->getFilterField($index, $filter->field)]['value'] = $filter->value, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
                 $filter instanceof Condition\GreaterThanCondition => $filterQueries[]['range'][$this->getFilterField($index, $filter->field)]['gt'] = $filter->value, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
--- a/packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
+++ b/packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
@@ -243,7 +243,7 @@ final class ElasticsearchSearcher implements SearcherInterface
         foreach ($filters as $filter) {
             match (true) {
                 $filter instanceof Condition\IdentifierCondition => $filterQueries[]['ids']['values'][] = $filter->identifier, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
-                $filter instanceof Condition\SearchCondition => $filterQueries[]['bool']['must']['query_string']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
+                $filter instanceof Condition\SearchCondition => $filterQueries[]['multi_match']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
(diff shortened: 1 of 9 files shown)


botzkobg:
All failing tests because of my changes are fixed now. The tests that still fail are the once that were failing before my changes.

alexander-schranz:
Thx for work on this. I need to deep read the elasticsearch docs, but first few you are right that `query_string` is not the expected behaviour here. Not happy with the tests, need to have a look at it and let you know.

But if I remember correctly this should also be fixed on Opensearch then?

botzkobg:
Hi @alexander-schranz 

I have fixed the Opensearch too, so it also uses `multi_match`. When you have a time to go over the PR let me know what you don't like about the tests and I will work on that. Thank you.

Kind regards.
@botzkobg

packages/seal-memory-adapter/src/MemorySearcher.php
```diff
@@ -164,7 +164,8 @@ private function filterDocuments(Index $index, array $documents, object $filter,
                 $searchableDocument = $this->getSearchableDocument($index->fields, $document);
 
                 $text = \json_encode($searchableDocument, \JSON_THROW_ON_ERROR);
-                $terms = \explode(' ', $filter->query);
+                $query = preg_replace('#(/|"|\\\\)#', '\\\\$0', $filter->query);
```
alexander-schranz: Would like to avoid the regex here and go with something like:


```suggestion
                $query = \trim(\json_encode($filter->query, \JSON_THROW_ON_ERROR)), '"');
```

would that work?

alexander-schranz (changes requested):
Still need check the other parts of my previous comments. https://github.com/PHP-CMSIG/search/pull/629#issuecomment-3723457958

Can no yet give you a time if I will able todo it this week or not.

packages/seal-memory-adapter/src/MemorySearcher.php
botzkobg: The point here is to escape `/`, `"` and `\`, because they will be escaped in the JSON string representing the document. For example `He said "Go" and made an arrow (/\) from sticks` converted to JSON will be `"He said \"Go\" and made an arrow (\/\\) from sticks"`. The suggested change is not correct because it doesn't escape the special characters. I can replace the `preg_replace()` with `str_replace()`

```
$query = str_replace(['\\', '/', '"'], ['\\\\', '\/', '\"'], $filter->query);
```

This will be much easier to read and understand.

Let me know what do you think about that.

alexander-schranz: but json_encode would handle all special character and so we not require handling all special characters our selves.

botzkobg: Ahh you are correct. I missed the `json_encode()` inside the `trim()` 🤦‍♂️

packages/seal-elasticsearch-adapter/src/ElasticsearchSearcher.php
```diff
@@ -243,7 +243,7 @@ private function recursiveResolveFilterConditions(Index $index, array $filters,
         foreach ($filters as $filter) {
             match (true) {
                 $filter instanceof Condition\IdentifierCondition => $filterQueries[]['ids']['values'][] = $filter->identifier, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
-                $filter instanceof Condition\SearchCondition => $filterQueries[]['bool']['must']['query_string']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
+                $filter instanceof Condition\SearchCondition => $filterQueries[]['multi_match']['query'] = $filter->query, // @phpstan-ignore-line offsetAccess.nonOffsetAccessible
```
alexander-schranz: Why not part of the bool anymore?

```php
$filterQueries[]['bool']['must']['multi_match']['query']
```

did that make any unexpected problems?

botzkobg: I didn't think to leave it under `bool.must`. I just made some tests and there doesn't seem to be a problems, so I will move it there.

alexander-schranz:
@botzkobg thank you for your effort and fixing this 💪  . I will take a look at the tests when I have more time, currently happy to merge this to get the bug fixed.

If you have some time I'm always curious how people are using Search Engines and/or SEAL so if you got some minutes leave a comment in following discussions:

 - https://github.com/PHP-CMSIG/search/discussions/416
 - https://github.com/PHP-CMSIG/search/discussions/457

botzkobg:
@alexander-schranz , I'm sorry to bother you  but do you have any idea when you will release a version with this fix?

alexander-schranz:
Was hoping that we could get https://github.com/PHP-CMSIG/search/pull/632 also in it but looks like nothing goes forwarded there will do a release a soon as possible for me.

alexander-schranz:
Release in 0.12.7 Thank you!

PR #632, opened 2026-01-10
Add support for Symfony 8 for Symfony and Yii Integration

baldurrensch:


alexander-schranz:
Thx for the pull request this also requires to update the `.examples/symfony` to allow there also Symfony 8 (we keep the old version as it is also used for tests so normally no other changes as the composer.json/lock is required).

Also we need to check where else Symfony components are used. Think Yii and Spiral Integration also supports Symfony 8 already and could be updated.

PR #644, opened 2026-01-31
Fix type casting for --bulk-size option in Symfony command

### LINKED ISSUE #643: cmsig:seal:reindex --bulk-size causes type casting error in symfony bundle
When running the `cmsig:seal:reindex` command with the `--bulk-size` option, the command fails with a type casting error.

**Steps to Reproduce**
1. Run the command:
   bin/console cmsig:seal:reindex --bulk-size=100
   -> Fails with a casting/type error

**Expected Behavior**
The `--bulk-size` option should be parsed as an integer.

**Actual Behavior**
The command throws a type casting error as soon as `--bulk-size` is provided.

**Environment**
- Symfony: 7.x
- PHP: 8.x

**Possible Cause**
It appears the `--bulk-size` option is being passed as a string instead of being cast to an int before being used.

**Example fix:**
$bulkSize = (int) $input->getOption('bulk-size');

mario-fehr:
Fixes #643


mario-fehr:
@alexander-schranz can you give me a hint about the pipeline erros?

alexander-schranz:
Can be ignored. PhPstan is a bug in Symfony. Spiral is a little bit flacky currently.

alexander-schranz:
@mario-fehr Thank you!

PR #648, opened 2026-02-11
Add refresh flag when return_slow_promise_result is enabled.

### LINKED ISSUE #645: ElasticsearchIndexer::bulk missing refresh support when `return_slow_promise_result` is enabled
### Problem

In the Elasticsearch adapter, `ElasticsearchIndexer::save()` and `ElasticsearchIndexer::delete()` support the `return_slow_promise_result` option by setting the Elasticsearch `refresh` parameter. This ensures that documents are immediately visible (or removed) in subsequent `/_search` calls, which is especially useful for PHPUnit / functional tests.

However, `ElasticsearchIndexer::bulk()` currently does NOT support any refresh behavior at all. There is no way to make bulk-indexed or bulk-deleted documents immediately searchable when `return_slow_promise_result` is enabled.

This makes `bulk()` inconsistent with `save()` / `delete()` and hard to use in test environments.

### Expected

When `return_slow_promise_result` is enabled, `bulk()` should pass a `refresh` parameter to Elasticsearch (similar to `save()` / `delete()`), so that:

- Bulk-indexed documents are immediately available via `/_search`
- Bulk-deleted documents are immediately removed from search results
- `bulk()` can be reliably used in PHPUnit / functional tests

### Proposed change

Add refresh handling to `bulk()` when `return_slow_promise_result` is enabled, similar to `save()` / `delete()`.

Example:

```
$params = [
'body' => [],
'refresh' => ($options['return_slow_promise_result'] ?? false) ? 'wait_for' : 'false',
];
```
`wait_for` is good enough for phpUnit tests and not so costly as `true`

### Additional Context

Happy to open a PR with the minimal change.

mario-fehr:
Fixes #645

### DIFF
--- a/packages/seal-elasticsearch-adapter/src/ElasticsearchIndexer.php
+++ b/packages/seal-elasticsearch-adapter/src/ElasticsearchIndexer.php
@@ -100,7 +100,11 @@ final class ElasticsearchIndexer implements IndexerInterface
 
         $batchIndexingResponses = [];
         foreach (BulkHelper::splitBulk($saveDocuments, $bulkSize) as $bulkSaveDocuments) {
-            $params = ['body' => []];
+            $params = [
+                'body' => [],
+                // TODO refresh should be refactored with async tasks
+                'refresh' => ($options['return_slow_promise_result'] ?? false) ? 'true' : 'false', // update document immediately, so it is available in the `/_search` api directly
+            ];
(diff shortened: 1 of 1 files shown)


alexander-schranz (approved):

alexander-schranz:
@mario-fehr thank you!

PR #649, opened 2026-02-12
Add support for Elasticsearch via SSL Proxy

### LINKED ISSUE #638: Support Elasticsearch via SSL Proxy
My Elasticsearch Host ist behind a SSL-Proxy like https://my-es.my-domain.com:443/ with authentication active.

If i set SEAL_DSN=elasticsearch://user:password@my-es.my-domain.com:443 i get "no alive nodes found".
If i prefix the host with "https://" in ClientBuilder.php it works.

So there should be a way to define if the host needs that https prefix.
Very basic check should be if the port is 443, but it would be even better to manually force that behavior.

mario-fehr:
Fix #638

### DIFF
--- a/packages/seal-elasticsearch-adapter/src/ElasticsearchAdapterFactory.php
+++ b/packages/seal-elasticsearch-adapter/src/ElasticsearchAdapterFactory.php
@@ -45,6 +45,7 @@ class ElasticsearchAdapterFactory implements AdapterFactoryInterface
      *     port?: int,
      *     user?: string,
      *     pass?: string,
+     *     query: array<string, string>,
      * } $dsn
      */
     public function createClient(array $dsn): Client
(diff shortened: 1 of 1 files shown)


packages/seal-elasticsearch-adapter/src/ElasticsearchAdapterFactory.php
```diff
@@ -45,6 +45,7 @@ public function createAdapter(array $dsn): AdapterInterface
      *     port?: int,
      *     user?: string,
      *     pass?: string,
+     *     query: array<string, string>,
```
alexander-schranz:
```suggestion
     *    query: array<string, string|string[]>,
```

packages/seal-elasticsearch-adapter/src/ElasticsearchAdapterFactory.php
```diff
@@ -59,8 +60,12 @@ public function createClient(array $dsn): Client
             return $client;
         }
 
+        $useTls = ($dsn['query']['tls'] ?? false) === true;
```
alexander-schranz: Query can be also assocative array so we need check that, and validate it:

```suggestion
        $tlsQuery = $dsn['query']['tls'] ?? 'false';
        \assert(\is_string($tlsQuery), 'The "tls" query param must be a string.');
        $useTols = \filter_var($tlsQuery, \FILTER_VALIDATE_BOOL, \FILTER_REQUIRE_SCALAR);
```

alexander-schranz (changes requested):
Small change request to best consistent with query being `array<string, string|string[]>` in our case.

alexander-schranz (approved):

alexander-schranz:
@mario-fehr thx, for the OpensearchAdapterFactory it should be the same if you want to tackle that also.

PR #651, opened 2026-02-19
Add support for TLS with Typesense adapter

RamonNijman:
## Problem

The Typesense adapter currently only supports HTTP connections. This makes it
impossible to connect to any Typesense instance that requires HTTPS, such as:

- **Typesense Cloud** — which exclusively uses HTTPS on port 443
  (e.g., `xxx.a1.typesense.net:443`)
- **Self-hosted instances** behind a reverse proxy with SSL termination
  (e.g., Cloudflare Tunnel, Nginx with TLS, Traefik)

Without HTTPS support, users are unable to use SEAL with Typesense Cloud at all,
and self-hosted setups require insecure HTTP connections.

## Solution

Add support for a `tls=true` query parameter in the Typesense DSN to enable
HTTPS connections.

## Solution

Add support for a `tls=true` query parameter in the Typesense DSN to enable
HTTPS connections.

### Usage

HTTP
SEAL_DSN="typesense://S3CR3T@127.0.0.1"

HTTPS
SEAL_DSN="typesense://S3CR3T@127.0.0.1?tls=true"
SEAL_DSN="typesense://S3CR3T@127.0.0.1:443?tls=true"

### Changes
- `TypesenseAdapterFactory`: Parse `tls` query parameter from DSN and set
  protocol to `https` when `tls=true`
- Updated documentation

### Backwards Compatible
Yes — existing `typesense://` DSN strings without `?tls=true` continue to
work unchanged with HTTP.

### DIFF
--- a/packages/seal-typesense-adapter/src/TypesenseAdapterFactory.php
+++ b/packages/seal-typesense-adapter/src/TypesenseAdapterFactory.php
@@ -44,6 +44,7 @@ class TypesenseAdapterFactory implements AdapterFactoryInterface
      *     host: string,
      *     port?: int,
      *     user?: string,
+     *     query: array<string, string>,
      * } $dsn
      */
     public function createClient(array $dsn): Client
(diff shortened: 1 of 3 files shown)


packages/seal-typesense-adapter/src/TypesenseAdapterFactory.php
```diff
@@ -58,14 +59,20 @@ public function createClient(array $dsn): Client
             return $client;
         }
 
+        $tlsQuery = $dsn['query']['tls'] ?? 'false';
+        \assert(\is_string($tlsQuery), 'The "tls" query param must be a string.');
+        $useTls = \filter_var($tlsQuery, \FILTER_VALIDATE_BOOL, \FILTER_REQUIRE_SCALAR);
+        $protocol = $useTls ? 'https' : 'http';
+        $port = $dsn['port'] ?? $useTls ? 443 : 8108;
```
alexander-schranz: This does not work as expected as it ` $dsn['port'] ?? $useTls ? 443 : 8108;` = `($dsn['port'] ?? $useTls) ? 443 : 8108;` and so custom ports never used.


```suggestion
        $port = ($dsn['port'] ?? $useTls) ? 443 : 8108;
```

If we change the port depending on `useTls` we would require do the same for Elasticsearch Adapter.

alexander-schranz (changes requested):
Thx for the pull request. One issue:

packages/seal-typesense-adapter/src/TypesenseAdapterFactory.php
RamonNijman: I've implemented a fix for the fallback scenario.

I've also implemented a change for Elasticsearch to also use port 443 when TLS is enabled.

alexander-schranz: Thx

PR #662, opened 2026-03-01
Add possibility for reusable ReindexProvider classes (BC Break ⚠️) 

alexander-schranz:
Currently it is not possible to create a reusable ReindexProvider as all ReindexProviders require to implement a static method `public static function getIndex(): string` we should get rid of them so a people can ReindexProviders which are reusable and just configurable by there constructor.

## BC Breaks

The `ReindexProviderInterface` and `getIndex` method is not longer static:

```diff
class YourClass implements ReindexProviderInterface {
    // ...
    
-   public static function getIndex(): string
+   public function getIndex(): string
    {
        return 'test';
    }
}
```

### DIFF
--- a/.examples/laravel/app/Search/BlogReindexProvider.php
+++ b/.examples/laravel/app/Search/BlogReindexProvider.php
@@ -35,7 +35,7 @@ class BlogReindexProvider implements ReindexProviderInterface
         ];
     }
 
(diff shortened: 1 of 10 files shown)


alexander-schranz:
replaced by https://github.com/PHP-CMSIG/search/pull/666

PR #666, opened 2026-03-02
Add possibility for reusable ReindexProvider classes (BC Break ⚠️)

### LINKED ISSUE #615: Make ReindexProvider getIndex none static
The idea behind that we did make the method of the ReindexProvider static:

```php
    public static function getIndex(): string;
```

was that static methods are able to be called without instantiate the class trigger lazy service to be loaded.

The problem is if we want reuse a ReindexProvider for example when creating a Index per language we would require to create a Reindex class also per language. As languages are in Sulu depending on other service it would be best to make it none static so something like:

```php
    public function getIndex(): string
    {
        return 'website' . $this->locale;
    }
```

Can be implemented.

Sadly this would be kind of BC break: https://3v4l.org/nfFFJ not sure why but PHP does not allow that a implementation does make a method of interface static.

@Toflar think you also had in contao once request similar change? But as the name != real index name you could workaround it.

alexander-schranz:
Currently it is not possible to create a reusable ReindexProvider as all ReindexProviders require to implement a static method `public static function getIndex(): string` we should get rid of them so people can create ReindexProviders which are reusable and just configurable by there constructor.

This also add possibility to wrap / decorate ReindexProviders easier, which we require for ODM (#81 #661).

## BC Breaks

The `ReindexProviderInterface` and `getIndex` method is not longer static and the interface was renamed to `StaticReindexProviderInterface` as in a `DynamicReindexProviderInterface` without getIndexName method will be added in upcoming PR:

```diff
-class YourClass implements ReindexProviderInterface {
+class YourClass implements StaticReindexProviderInterface {
    // ...
    
-   public static function getIndex(): string
+   public function getIndexName(): string
    {
        return 'test';
    }
}
```

If you need to support 0.12 and 0.13 the same time you can implement both methods and keep using the deprecated ReindexProviderInterface until 0.14 or 1.0:

```diff
class YourClass implements ReindexProviderInterface {
    // ...
    
    public static function getIndex(): string
    {
        return 'index';
    }
    
+   public function getIndexName(): string
+   {
+        return self::getIndex();
+   }
}
```

fixes #615

### DIFF
--- a/.examples/laravel/app/Search/BlogReindexProvider.php
+++ b/.examples/laravel/app/Search/BlogReindexProvider.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace App\Search;
 
 use CmsIg\Seal\Reindex\ReindexConfig;
-use CmsIg\Seal\Reindex\ReindexProviderInterface;
+use CmsIg\Seal\Reindex\StaticReindexProviderInterface;
 
-class BlogReindexProvider implements ReindexProviderInterface
+class BlogReindexProvider implements StaticReindexProviderInterface
 {
     public function total(): int|null
     {
--- a/.examples/laravel/app/Search/BlogReindexProvider.php
+++ b/.examples/laravel/app/Search/BlogReindexProvider.php
@@ -35,7 +35,7 @@ class BlogReindexProvider implements ReindexProviderInterface
         ];
     }
 
-    public function getIndex(): string
+    public function getIndexName(): string
     {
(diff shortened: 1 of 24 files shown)


Toflar:
What happened with the idea outlined in https://github.com/PHP-CMSIG/search/issues/615#issuecomment-3495559467? Not good anymore? 😊

alexander-schranz:
@Toflar thx for the hint forgot about that :)

PR #674, opened 2026-03-24
Fix filtering on TextField for Solr adapter

### LINKED ISSUE #671: [Solr] Filtering on TextField with searchable:false throws "undefined field" error
When a `TextField` is configured with `searchable: false` but `filterable: true`, the Solr adapter creates the field in the schema without a `.raw` subfield. However, when building search queries with filters on that field, `SolrSearcher::getFilterField()` always returns `{field}.raw` for any `TextField`, causing Solr to throw an "undefined field" error.

**Steps to reproduce**

1. Define an index with a `TextField` that is not searchable but filterable:

```php
use CmsIg\Seal\Schema\Field;

new Field\TextField('status', searchable: false, filterable: true)
```

2. Create the index in Solr (schema is generated correctly without `.raw` subfield)

3. Execute a search with a filter on that field:

```php
$results = $searcher->search(
    $index->createSearchBuilder()
        ->addFilter('status', 'active')
        ->getSearch()
);
```

4. Solr returns error: `undefined field status.raw`

I am not sure about solution, but I think that the checking in the `SolrSearcher` should be the same with `SolrSchemaManager`

```php
private function getFilterField(Index $index, string $name): string
{
    $field = $index->getFieldByPath($name);

    if ($field instanceof Field\TextField && $field->searchable && ($field->filterable || $field->sortable || $field->facet)) {
        return $name . '.raw';
    }

    return $name;
}
```

I can prepare a PR if required.

danilovq:
This PR fixes issue #671 where filtering on a TextField with `searchable: false` and `filterable: true` causes an "undefined field" error in Solr.

### Problem
- `SolrSchemaManager` only creates `.raw` subfield when TextField is searchable AND filterable/sortable/facet
- `SolrSearcher::getFilterField()` always returned `.raw` for any TextField being filtered
- Mismatch causes query builder to reference non-existent `.raw` field

### Solution
Update `getFilterField()` to use the same condition as schema creation:
- Only append `.raw` when field is a TextField AND searchable AND (filterable OR sortable OR facet)

Fixes #671

### DIFF
--- a/packages/seal-solr-adapter/src/SolrSearcher.php
+++ b/packages/seal-solr-adapter/src/SolrSearcher.php
@@ -224,7 +224,7 @@ final class SolrSearcher implements SearcherInterface
     {
         $field = $index->getFieldByPath($name);
 
-        if ($field instanceof Field\TextField) {
+        if ($field instanceof Field\TextField && $field->searchable && ($field->filterable || $field->sortable || $field->facet)) {
             return $name . '.raw';
         }
 


alexander-schranz:
@danilovq thank you. This will be part of the next release. If you maybe still got some minutes I'm always interested who is using SEAL and how and also what type of Search Engiens they prefer. Please feel free to add some comment to following discussion:

 - [Are you working with SEAL ❓  Let us know!](https://github.com/PHP-CMSIG/search/discussions/416)
 - [Which Search Engines do you use and why ❓ ](https://github.com/PHP-CMSIG/search/discussions/457)

alexander-schranz:
Released in 0.12.10: https://github.com/PHP-CMSIG/search/releases/tag/0.12.10

PR #681, opened 2026-04-22
Fix not dump config files for in-memory loupe setups

Toflar:
Noticed while working on some functional tests ^^

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeHelper.php
+++ b/packages/seal-loupe-adapter/src/LoupeHelper.php
@@ -28,7 +28,7 @@ final class LoupeHelper
     /**
      * @var array<string, Configuration>
      */
-    private static array $inMemoryConfigurations = [];
+    private array $inMemoryConfigurations = [];
 
     /**
      * @var array<string, Loupe>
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@Toflar That make sense, but we should still store the configuration in memory (static PHP variable and read from it) to simulate the same behaviour as when file persistent is used.

Toflar:
Erm, why? I think we definitely shouldn't.

alexander-schranz:
@alexander-schranz keep the behaviour the same. SEAL will always use the `createIndex` configuration, if configuration is changed `createIndex` need be called to update it. It avoids that somebody can change index configuration without explicit call create index.

SEAL is here a lot stricter as Loupe itself.

Toflar:
Like so?

alexander-schranz:
👍 Looks good, will test it, Lint failing seems unrelated, so you can ignore it.

Toflar:
Could you try to release this with the next bugfix release? It keeps freaking me out in my functional tests 🤣

alexander-schranz (approved):

alexander-schranz:
@Toflar I did a few adjustments can you test if all still works as expected for you?

Toflar:
LGTM!

PR #684, opened 2026-05-04
Added support for configuring Loupe via DSN

Toflar:


alexander-schranz:
Like the idea, do we need to make this index specific? `?configuration[blog]=...&?configuration[news]=...` ?

Toflar:
Why? The `$directory` is already a parameter - you cannot have multiple configurations per `$directory`?

alexander-schranz:
@Toflar SEAL was always multi Index, the directory is just the root directory. Multi indexes for the same DSN was always possible even before Loupe switched from .sqlite file to directory itself SEAL adapter already used a directory per index.

 - `var/indexes`
    - `<index_A_name>` Loupe Dir
    - `<index_B_name>` Loupe Dir

https://github.com/Toflar/cmsig-search/blob/6e1cc2fb402a8818eea88db4fde1bfc1a2fff582/packages/seal-loupe-adapter/src/LoupeHelper.php#L155-L158

Toflar:
Ah right. Hmm, but then the strings could get pretty long 😊 But I guess that shouldn't be an issue?

alexander-schranz:
@Toflar don't see any issue with it. We still could support both `?configuration=` and `?configuration[blog]`, but maybe we better give the `strings` as constructor parameter? The config is currently once configuration once created by SEAL and cached as SEAL not support runtime changes on indexes without call index create command.

Toflar:
Indeed. So something like this?

PR #689, opened 2026-05-22
Fix RediSearch Adapter Filters when field not searchable ( :warning: Require Reindex )

alexander-schranz:
As reported in Sulu the `locale` filter didn't work as expected as the field is `searchable: false` but `filterable: true` and so not correctly handled.

### Requires Reindex

Run: `cmsig:seal:reindex` command to make sure the index works as expected.

### DIFF
--- a/packages/seal-redisearch-adapter/src/RediSearchIndexer.php
+++ b/packages/seal-redisearch-adapter/src/RediSearchIndexer.php
@@ -29,7 +29,6 @@ final class RediSearchIndexer implements IndexerInterface
     ) {
         $this->marshaller = new Marshaller(
             dateFormat: 'U',
-            addRawFilterTextField: true,
             geoPointFieldConfig: [
                 'latitude' => 1,
                 'longitude' => 0,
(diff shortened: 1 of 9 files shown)


jbfewo:
:+1:

packages/seal-redisearch-adapter/src/RediSearchIndexer.php
```diff
@@ -29,7 +29,6 @@ public function __construct(
     ) {
         $this->marshaller = new Marshaller(
             dateFormat: 'U',
-            addRawFilterTextField: true,
```
alexander-schranz, line 32: Changing the marshaller config requires a drop reindex, this need be considered as a BC Break.

packages/seal-redisearch-adapter/src/RediSearchSchemaManager.php
```diff
@@ -147,7 +145,7 @@ private function createJsonFields(array $fields, string $prefix = '', string $js
                     ],
                 ] : [], $field->filterable || $field->facet || $field->sortable ? [
                     $name . '.raw' => [
-                        'jsonPath' => $jsonPathRaw,
+                        'jsonPath' => $jsonPath,
```
alexander-schranz, line 148: Changing the JSON path requires a drop Reindex, this need be considered as a BC Break.

jbfewo:
💐

alexander-schranz:
@jbfewo fix released in 0.12.12.

alexander-schranz:
@jbfewo if you have some time feel free to attend the following discussions:


  - <a href="https://github.com/PHP-CMSIG/search/discussions/416">Are you working with SEAL? Let us know!</a>
  
  - <a href="https://github.com/PHP-CMSIG/search/discussions/457">Which Search Engines do you use and why?</a>

PR #696, opened 2026-06-08
Increase PHP min Version to 8.2 (BC Break ⚠️)

alexander-schranz:
The 8.1 was because of incompatibility to the integrations not tested. To make sure that all versions are tested we now set the min verson to PHP 8.2.

### BC Breaks

If you are maintaining a library or project using SEAL which need to support PHP 8.1 you may also need increase your PHP Version or create a backwards compatibility layer to support older SEAL and newer SEAL Version at the same time.


jdreesen:
Why is it a BC break? It only raises the minimum PHP version :thinking:

alexander-schranz:
@jdreesen If you are using SEAL inside a library or project which supports multiple PHP Versions including PHP 8.1 you may need create a backwards compatibility layer to support older and newer SEAL version at the same time. So it may requires changes inside your project. That is why I mark such things as a BC Break.

PR #697, opened 2026-06-08
Replace deprecated opensearch-php bootstrap APIs in seal-opensearch-adapter

Lustmored:
## Summary

Replace deprecated `opensearch-php` APIs in `seal-opensearch-adapter` and switch client bootstrap to the PSR-based transport path.

## Changes

- bump `opensearch-project/opensearch-php` to `^2.4`
- replace deprecated `ClientBuilder` usage in runtime code
- replace deprecated `OpenSearch\Common\Exceptions\*` imports with the new exception namespace
- add `BaseUriRequestFactory` for DSN-based request creation on top of the PSR transport stack
- preserve DSN features used by the adapter, including `tls=true`, basic auth, and optional path prefixes
- route integration tests through `OpensearchAdapterFactory::createClient()`
- add unit coverage for:
  - `BaseUriRequestFactory`
  - DSN normalization in the OpenSearch test helper
- update OpenSearch docs/examples to remove deprecated `ClientBuilder` usage
- document the required standalone client/factory packages for Symfony and Guzzle examples

### DIFF
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -688,17 +688,90 @@ It requires an instance of the ``Adapter`` which we did install before to connec
 
                 Use the following code to create a new ``Engine`` using the ``Opensearch`` adapter:
 
+                First install the adapter:
+
+                .. code-block:: bash
+
+                    composer require cmsig/seal-opensearch-adapter
+
+                Via DSN
+                ~~~~~~~
+
+                When you use the adapter via DSN, ``opensearch-php`` uses discovery to find the installed PSR implementations automatically.
+
+                Your project needs:
+
+                - a ``psr/http-client-implementation``
+                - a ``psr/http-factory-implementation``
+
+                If your framework or application already provides them, no additional HTTP client package is needed.
+
+                Example DSNs:
+
+                .. code-block:: env
+
+                    opensearch://127.0.0.1:9200
+                    opensearch://127.0.0.1:9200?tls=true
+                    opensearch://username:password@127.0.0.1:9200?tls=true
+
+                Example standalone usage via DSN:
+
                 .. code-block:: php
 
                     <?php
 
-                    use OpenSearch\ClientBuilder;
-                    use CmsIg\Seal\Adapter\Opensearch\OpensearchAdapter;
+                    use CmsIg\Seal\Adapter\AdapterFactory;
+                    use CmsIg\Seal\Adapter\Opensearch\OpensearchAdapterFactory;
                     use CmsIg\Seal\Engine;
 
-                    $client = ClientBuilder::create()->setHosts([
-                        '127.0.0.1:9200'
-                    ])->build()
+                    $adapterFactory = new AdapterFactory([
+                        OpensearchAdapterFactory::getName() => new OpensearchAdapterFactory(),
+                    ]);
+
+                    $engine = new Engine(
+                        $adapterFactory->createAdapter('opensearch://127.0.0.1:9200?tls=true'),
+                        $schema,
+                    );
+
+                Manual client creation
+                ~~~~~~~~~~~~~~~~~~~~~~
+
+                If you do not use the adapter via DSN, create an ``OpenSearch\Client`` yourself and pass it to the adapter.
+
+                For example, you can use one of these PSR-18 / PSR-17 stacks:
+
+                .. code-block:: bash
+
+                    # Symfony HTTP Client
+                    composer require symfony/http-client nyholm/psr7
+
+                    # Guzzle
+                    composer require guzzlehttp/guzzle guzzlehttp/psr7
+
+                .. code-block:: php
+
+                    <?php
+
+                    use CmsIg\Seal\Adapter\Opensearch\OpensearchAdapter;
+                    use CmsIg\Seal\Engine;
+                    use OpenSearch\Client;
+
+                    /** @var Client $client */
+                    // Manual client creation example.
+                    //
+                    // Symfony example:
+                    // composer require symfony/http-client nyholm/psr7
+                    // use OpenSearch\SymfonyClientFactory;
+                    // $client = (new SymfonyClientFactory())->create([
+                    //     'base_uri' => 'http://127.0.0.1:9200',
+                    // ]);
+                    //
+                    // Guzzle example:
+                    // composer require guzzlehttp/guzzle guzzlehttp/psr7
+                    // use OpenSearch\GuzzleClientFactory;
+                    // $client = (new GuzzleClientFactory())->create([
+                    //     'base_uri' => 'http://127.0.0.1:9200',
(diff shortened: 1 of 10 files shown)


Lustmored:
Part of this PR is compatibility code for Opensearch-PHP 2.4.* for things that were fixed in later released (the most important being header format normalization). The best solution would be to require ^2.6, but that implies PHP >= 8.2, while this project sticks to 8.1, so I went with compatibility code instead of such hard dependency bump.

alexander-schranz:
@Lustmored thank you for your pull request. Here some quick questions:

> The best solution would be to require ^2.6, but that implies PHP >= 8.2, while this project sticks to 8.1, so I went with compatibility code instead of such hard dependency bump.

We should target the 0.13 then for this. That requires PHP 8.2 atleast.

>         .. code-block:: bash
>
>           # Symfony HTTP Client
>           composer require symfony/http-client nyholm/psr7
>
>           # Guzzle
>           composer require guzzlehttp/guzzle guzzlehttp/psr7

This feels strange does Opensearch not support PSR compatible clients so we could use the discovery feature of Http Client Implementation which selects the best HTTP Client based on the used frameworks and installed versions?

https://github.com/psr-discovery/http-client-implementations

I did not yet have deeper look at the code changes. Just want to discuss this first.

Lustmored:
> This feels strange does Opensearch not support PSR compatible clients so we could use the discovery feature of Http Client Implementation which selects the best HTTP Client based on the used frameworks and installed versions?
> 
> https://github.com/psr-discovery/http-client-implementations
> 
> I did not yet have deeper look at the code changes. Just want to discuss this first.

Opensearch-PHP provides factories for Guzzle and Symfony, but not for discovery for some reason. That's why I've included code relying on `Psr17FactoryDiscovery` in `OpensearchAdapterFactory`.

The probably misleading readme entries are here as I wanted to give user out of the box working examples. I think we could rewrite it so that the intent is more clear - user can use any configured PSR HTTP factories.

> We should target the 0.13 then for this. That requires PHP 8.2 atleast.

Good call! I will rebase it onto 0.13 and change the PR target

Lustmored:
I rebased and changed target of this PR to 0.13. I also bumped opensearch-php to `^2.6` and removed now unnecessary compatibility code and rewrote README to be more detailed regarding discovery and manual client building.

alexander-schranz:
@Lustmored I need check this out its a lot of unnecessary code if we look the previous version. I think best would to contribute again a PSR compatible Factory to Opensearch PHP to avoid that bootstrap a OpenSearch client takes such many lines. HTTP Client PSR was specially created to avoid things like Symfony Factory and Guzzle Factory it this is more huge step back in my opinion but I'm not in the loop.

packages/seal-opensearch-adapter/src/OpensearchAdapterFactory.php
```diff
@@ -59,24 +62,20 @@ public function createClient(array $dsn): Client
             return $client;
         }
 
-        $tlsQuery = $dsn['query']['tls'] ?? 'false';
-        \assert(\is_string($tlsQuery), 'The "tls" query param must be a string.');
-        $useTls = \filter_var($tlsQuery, \FILTER_VALIDATE_BOOLEAN, \FILTER_REQUIRE_SCALAR);
-        $scheme = $useTls ? 'https' : 'http';
-        $port = $dsn['port'] ?? ($useTls ? 443 : 9200);
```
alexander-schranz, line 66: keep this logic here.

packages/seal-opensearch-adapter/tests/ClientHelper.php
```diff
@@ -23,11 +24,67 @@ final class ClientHelper
...
+            $host = $_ENV['OPENSEARCH_HOST'] ?? '127.0.0.1:9200';
+            $host = \is_string($host) ? $host : '127.0.0.1:9200';
+
+            $adapterFactory = new AdapterFactory([
+                OpensearchAdapterFactory::getName() => new OpensearchAdapterFactory(),
+            ]);
+
+            self::$client = (new OpensearchAdapterFactory())->createClient(
+                $adapterFactory->parseDsn(self::normalizeDsn($host)),
+            );
```
alexander-schranz, line 36: Do not use factory here instead create the client the fastest way as possible like before, just using one of the new builder.

packages/seal-opensearch-adapter/src/OpensearchAdapterFactory.php
```diff
@@ -59,24 +62,20 @@ public function createClient(array $dsn): Client
...
+        $psrRequestFactory = Psr17FactoryDiscovery::findRequestFactory();
+        $streamFactory = Psr17FactoryDiscovery::findStreamFactory();
+        $uriFactory = Psr17FactoryDiscovery::findUriFactory();
 
-        $client = ClientBuilder::create()->setHosts([
-            $scheme . '://' . $dsn['host'] . ':' . $port,
-        ]);
```
alexander-schranz, line 70: to not readd a new service, instead keep it simple and just call one of the new builders. Think guzzle was used previously? So we should keep guzzle for this, if we need require it in the composer.json add it to composer.json.

packages/seal-opensearch-adapter/tests/ClientHelperTest.php
alexander-schranz, line 1: remove this.

packages/seal-opensearch-adapter/tests/BaseUriRequestFactoryTest.php
alexander-schranz, line 1: remove this.

packages/seal-opensearch-adapter/src/BaseUriRequestFactory.php
alexander-schranz, line 1: remove this.

docs/getting-started/index.rst
```diff
@@ -688,17 +688,90 @@ It requires an instance of the ``Adapter`` which we did install before to connec
...
+                    // $client = (new SymfonyClientFactory())->create([
+                    //     'base_uri' => 'http://127.0.0.1:9200',
+                    // ]);
+                    //
+                    // Guzzle example:
+                    // composer require guzzlehttp/guzzle guzzlehttp/psr7
+                    // use OpenSearch\GuzzleClientFactory;
+                    // $client = (new GuzzleClientFactory())->create([
+                    //     'base_uri' => 'http://127.0.0.1:9200',
+                    // ]);
```
alexander-schranz, line 774: TOOO MUCH infos here, keep docs simple, we are here in a getting started, who want to deep dive how to create the client differently will already look at the opensearch docs nothing which need live here in the SEAL docs.

alexander-schranz (changes requested):
Thx for the PR we should simplify this a lot more and not overcomplicate here things.

PR #698, opened 2026-06-16
Fix highlighting configuration without query

Toflar:
Currently, when searching without a query but e.g. filters and sorting (so a valid request), this leads to an issue if you also provide the fields to highlight configuration (at least for the Loupe adapter).
Why is that? Why not just ignore it? In most cases, the fields to highlight configuration is static and the current logic forces devs to first check if there is a query and if not, not set the fields to highlight?

Can't we just ignore that? Or automatically unset the fields to highlight setting if there is no query?

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeSearcher.php
+++ b/packages/seal-loupe-adapter/src/LoupeSearcher.php
@@ -158,14 +158,7 @@ final class LoupeSearcher implements SearcherInterface
             );
 
             foreach ($highlightFields as $highlightField) {
-                \assert(
-                    isset($hit['_formatted'])
-                    && \is_array($hit['_formatted'])
-                    && isset($hit['_formatted'][$highlightField]),
-                    \sprintf('Expected highlight field "%s" to be available in the search hit.', $highlightField),
-                );
-
-                $value = $hit['_formatted'][$highlightField];
+                $value = $hit['_formatted'][$highlightField] ?? null;
 
                 if (!\is_string($value)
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@Toflar I see, I currently expected as soon as highlighting fields are configured Loupe would return a `_formatted` field for that even that return is null if there is nothing to highlighted for that field.

I added the `assert` so I directly get error if Loupe changes something e.g. `_formatted` moved somewhere else in a new release.

But if you don't return null for configured highlighted fields in Loupe I'm fine todo that in SEAL and as you did remove the assert.

Fixing the PHPStan Lint issues I would invert the if statement to something like:

```php
                $value = null;
                if (\is_array($hit['_formatted'] ?? null)
                    && \is_string($hit['_formatted'][$highlightField] ?? null)
                    && \str_contains($hit['_formatted'][$highlightField], $highlightPreTag)
                ) {
                    $value = $hit['_formatted'][$highlightField];
                }

                $document['_formatted'][$highlightField] = $value;
```

I sure can have a look at Redis ~and Solr~ this week if you have no time for it which seems have a similar issue as the test fails.

Toflar:
You're right. I noticed why this is happening. It happens when you want to highlight fields that you did not configure to retrieve: https://github.com/loupe-php/loupe/pull/322

So fixing in Loupe :)

alexander-schranz:
@Toflar nice 👍 I still will check why Redis failed and maybe copy your test to the core :)

PR #701, opened 2026-06-21
Fix Redis and MemorySearcher with space behind and modify levenshtein algorithm

### LINKED ISSUE #700: [Redisearch] space at the end of query leads to Redis syntax error
It might be related to [#693](https://github.com/PHP-CMSIG/search/issues/693): if a search query ends with a space a RedisException is thrown with "SEARCH_SYNTAX Syntax error at offset..."

Affected Version 0.12.14

alexander-schranz:
Currently a space at end of string did error in RediSearch as it ended in something like`'%%%%'` which Redisearch don't like.

Also in MemorySearch the same term like e.g. `"Other Thing "` (with space at end) did end in all match. This is fixed in this Pull Request.

fixes #700

### DIFF
--- a/packages/seal-redisearch-adapter/src/RediSearchSearcher.php
+++ b/packages/seal-redisearch-adapter/src/RediSearchSearcher.php
@@ -273,7 +273,7 @@ final class RediSearchSearcher implements SearcherInterface
                         $escapedTerm = $this->escapeFilterValue($term);
 
                         // levenshtein algorithm per word length
-                        return match (\strlen($escapedTerm)) {
+                        return match (\strlen($term)) {
                             0, 1 => $escapedTerm,
                             2 => '%' . $escapedTerm . '%',
                             default => '%%' . $escapedTerm . '%%',
--- a/packages/seal-redisearch-adapter/src/RediSearchSearcher.php
+++ b/packages/seal-redisearch-adapter/src/RediSearchSearcher.php
@@ -268,10 +268,19 @@ final class RediSearchSearcher implements SearcherInterface
             };
 
             match (true) {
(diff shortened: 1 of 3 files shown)


packages/seal-redisearch-adapter/src/RediSearchSearcher.php
```diff
@@ -268,10 +268,19 @@ private function recursiveResolveFilterConditions(Index $index, array $condition
...
+
+                        // levenshtein algorithm per word length
+                        return match(\strlen($escapedTerm)) {
+                            0, 1 => $escapedTerm,
+                            2 => '%' . $escapedTerm . '%',
+                            default => '%%' . $escapedTerm . '%%',
+                        };
+                    },
+                    \array_filter(\explode(' ', $filter->query), \trim(...)),
+                )),
```
alexander-schranz, line 283: @jbfewo I also adopted the levenshtein so shorter terms not use the same levenshtein / fuzziness as longer words. Let me know what you think about it.

jbfewo: I like the idea of a length-dependent Levenshtein distance, but my gut feeling is that I would make it dependent on the length of the (unescaped) term.

alexander-schranz: Good catch 👍 will adopt it.

jbfewo: I'm playing around with that change and it definitively fixes the issue, even for a leading space ;)

I can't estimate how useful levenshtein=2 for a three letter word is. The importance might vary by language.

Issue #699, opened 2026-06-17
[Redisearch] Highlight can error without a search query


alexander-schranz:
Add a regression test for: https://github.com/PHP-CMSIG/search/pull/698 and check why Redis failed there even the issue was already resolved upstream in Loupe, it seems Redis running into similar issue which we need to tackle.

alexander-schranz:
false positive Redis has currently not yet support for Highlight feature see #491

PR #702, opened 2026-06-21
Fix highlight without search query in Loupe

### LINKED ISSUE #549: If index changes and reindexing is in progress highlighted fields do not work
When I change the index configuration (e.g. add a field) and the documents are still being reindexed, at least in Loupe everything works fine (documents can be found, some partially, others with the new field) unless I highlight the new field.
In this case I get "Expected highlight field to be set.".

Would it be possible to be more soft on highlighting? E.g. allow to configure to allow to ignore missing highlighted fields?

Otherwise, what would be a proper solution?

alexander-schranz:
It is common that the highlight is defined always even there is no search query submitted. Currently ~Redis and~ Loupe may have issues, while Loupe issue was already fixed upstream in 0.13.19 we maybe should be less strict and relax that part. 

fixes #549, see also #698, #699

### DIFF
--- a/packages/seal-loupe-adapter/src/LoupeSearcher.php
+++ b/packages/seal-loupe-adapter/src/LoupeSearcher.php
@@ -158,19 +158,12 @@ final class LoupeSearcher implements SearcherInterface
             );
 
             foreach ($highlightFields as $highlightField) {
-                \assert(
-                    isset($hit['_formatted'])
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Redis was a false positive, Redis currently has no highlighting implemented and so will not return anything here yet. So closing #699

PR #706, opened 2026-07-22
Add a static method to convert html to search-friendly text

### LINKED ISSUE #704: Helper Function to convert HTML to Text
Converting HTML to Text is not as easy as it seems as strip_tags of `<p>Test</p><p>Asdf</p>` ends in `TestAsdf` so maybe a Helper function would be best.

Maybe something in this direction:

```php
$html = preg_replace('~</?(?:p|div|br|th|td|li|h[1-6])\b[^>]*>~i', ' ', $html);
```

alexander-schranz:
It is very common that PHP Web Applications has some HTML in them. To avoid that we put HTML into the search engines we recommend to strip all html tags away and decode the html characters back to raw text.

## New API:

```php
$text = \CmsIg\Seal\Converter\HtmlToTextConverter::convert($html);
```

fixes #704

### DIFF
--- a/packages/seal/tests/Converter/HtmlToTextConverterTest.php
+++ b/packages/seal/tests/Converter/HtmlToTextConverterTest.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the CMS-IG SEAL project.
+ *
+ * (c) Alexander Schranz <alexander@sulu.io>
+ *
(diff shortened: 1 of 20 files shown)


packages/seal/tests/Converter/HtmlToTextConverterTest.php
alexander-schranz, line 1: @hual7 do you want to have a look if we catched any edge cases you could think of?

PR #720, opened 2026-08-06
Improve documentation for Identifier field type

AlwaysVictorious:
Clarified the description of the Identifier field type and its requirements.

### DIFF
--- a/docs/schema/index.rst
+++ b/docs/schema/index.rst
@@ -75,12 +75,12 @@ via ``sortable``, ``multiple``, ``filterable``, ``searchable`` and ``distinct`` 
 IdentifierField
 ~~~~~~~~~~~~~~~
 
-The ``Identifier`` field type is a ``Text`` field type that is used to
+The ``Identifier`` field type is used to
 identify a document in the Index. Its value is a PHP ``string``.
 
 Unlike ``Text`` and the other field types, it cannot be nullable. Every Index must have exactly one ``Identifier``. 
 
-The ``Identifier`` field type uses the ``Text`` field type's defaults as its options. The ``Identifier``'s options cannot be changed and are therefore the same for every Index.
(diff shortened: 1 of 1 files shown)


alexander-schranz (changes requested):
The changed text is not correct as Identifiers has its own defaults. So the last sentence should not be changed to reference the TextField type as its not true.

alexander-schranz:
PS: Thx for the effort, but it would help if you would create a pull request containing all your suggested docs changes instead per file, as we can then do a combined review instead have todo multiple reviews.

AlwaysVictorious:
I had to reread the documentation and now I see that there is a "options" section indeed. The problem I'm having (and that I will edit sometime soon but probably not today) is with "The Identifier field type is a special Text field type." This suggests extension hence there should be at least as many options to the Identifier field  as to the Text field. I quickly looked for a table, didn't see any, read "The defaults can not be changed and so are same for every index." and thought that hence it would use the Text field defaults.

PR 721 addresses as part of it a similar issue.  I will change this, change the offending sentence and will an options table here so that it matches the rest of the options sections. I will do this tomorrow I think

AlwaysVictorious:
> PS: Thx for the effort, but it would help if you would create a pull request containing all your suggested docs changes instead per file, as we can then do a combined review instead have todo multiple reviews.

To be honest, I did not plan on doing any of this. I just try to read the documentation and then when something can be improved, I do that expecting it to be one minor thing. And then there is another minor thing. And another.  From now on, if I encounter more changes, I will bundle them. There is however one PR I already made (#721)

AlwaysVictorious (commented):
Feel free to ask for other changes

alexander-schranz (approved):

alexander-schranz:
@AlwaysVictorious thank you!

